PR #7, opened 2021-02-06
Setup React Template

### LINKED ISSUE #3: Setup React template
With TypeScript, TailWind, Apollo Client and other needed dependencies.

blauks:
Closes #3


public/manifest.json
```diff
@@ -0,0 +1,25 @@
+{
```
FredrikAugust, line 1: If these files don't exist this file can probably be deleted to prevent 404 errors in the web console.

PR #9, opened 2021-02-07
Set up storybooks

### LINKED ISSUE #6: Setup Storybooks
Depends on #3 

[Storybooks](https://storybook.js.org/)

blauks:
Closes #6


FredrikAugust:
There is a lot of template code here that probably shouldn't be part of the finished PR. Of all the 22 files changed, only a few are relevant.

blauks:
True. I will remove all the unnecessary parts

FredrikAugust (changes requested):
Ref. comment

FredrikAugust (approved):
Is the `EMPTY` file for keeping the folder tracked in git? If so, the standard is to use an empty file called `.keep` or `.gitkeep`; https://dev.to/n350071/git-keep-empty-directory-by-gitkeep-aj1.

No need to change this now, but can be useful to know.

-------

You may wish to document the setup in the Wiki so it's easier for your teammates in @IT2901-SINTEF01/frontend to contribute.

Useful points:
- File name format. With this configuration, they have to be (in effect) `*.stories.tsx`
- How does a story look? Perhaps provide an example
- Should you be able to test [container components, or only presenational](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0)? If you should be able to test container components, it might be useful to provide a [wrapper](https://storybook.js.org/docs/react/writing-stories/decorators#context-for-mocking).

To keep things moving, I propose merging this PR and creating issues with the aforementioned points – if you concur.

blauks:
Great! Did not know about the .keep filename standard, but I will keep that in mind for next time. I will update the issues with your suggestions!:)

PR #12, opened 2021-02-07
Add build to CI workflow

### LINKED ISSUE #11: Add Github action workflow for building project
Add a workflow to install packages and build project

blauks:
Closes #11 

After stupidly many iterations, I managed to check if the project builds in the CI file.


.github/workflows/ci.yml
```diff
@@ -1,4 +1,4 @@
-name: Lint
+name: CI
```
FredrikAugust: Might be helpful to give it a useful name

.github/workflows/ci.yml
```diff
@@ -19,11 +19,14 @@ jobs:
...
+      
       - name: Run linters
         uses: wearerequired/lint-action@v1
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           # Enable linters
           eslint: true
           prettier: true
+
+      - name: Build project
```
FredrikAugust: Should this be a part of the `run-linters` job?

.github/workflows/ci.yml
blauks: Good point, I will find a better name

.github/workflows/ci.yml
blauks: Hehe:) Might have "koked" a little bit too much

FredrikAugust:
You now have two workflows that both perform a lot of the same setup. This means spinning up two virtual machines which check out your project and install dependencies.

This takes an unnecessary amount of time, and they should probably be merged into one workflow, two jobs.

I would recommend checkout out the documentation and understanding how the Actions work.

[It might also be relevant to setup a dependency between the two jobs.](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idneeds) If the build fails, is there any reason to test? Can you run a test with a project that doesn't build? These questions are left as exercises to the reader.

FredrikAugust:
With a dependency chain between the `build` and `test` jobs, do you need to `yarn` and checkout in both jobs, or can you omit the duplication?

FredrikAugust:
Might help you avoid having to resolve deps twice:

https://stackoverflow.com/questions/57498605/github-actions-share-workspace-artifacts-between-jobs

blauks:
Couldn't we just have one workflow and one job? This would make the yml file shorter and more easy to read.

FredrikAugust:
The reasoning for having two jobs was that it is more canonical as it performs two separate actions – builds an optimised build of the project (and install deps) and then runs linters.

Once you implement e2e tests with Cypress that will have to be it's own job, as merging it with linting (and in your proposal, building) would seem strange. Furthermore, testing can pass even though linting doesn't.

My proposed structure:
- Build (also install deps)
- Lint & Test (both should depend on build and use the artifacts from the job)

> This would make the yml file shorter and more easy to read.

It would be shorter, yes, but I do not agree that it would be easier to read. The complexity remains more or less the same, but I'd go as far as argue it's less canonical.

FredrikAugust:
For sharing artifacts between job, consider looking into `actions/cache`: https://github.community/t/artifcats-vs-cache-for-sharing-node-modules-between-jobs/16997

blauks:
According to the node-setup repository it will automatically cache Node specific files, so we do not have to do it ourselves.

FredrikAugust:
Looks good. Have you run the pipeline a few times to ensure the caching works as intended? You can check this in the action logs.

FredrikAugust:
If you've confirmed that it works feel free to merge.

FredrikAugust (approved):

blauks:
It works like a mountain

PR #16, opened 2021-02-08
Fix bug with lock hashing

blauks:


FredrikAugust:
Sorry, but it's still not right.

The intention behind the hashing is to use it to identify the exact resolution of all the packages. `package.json` doesn't include this. `package-lock.json` and `yarn.lock` do.

Read about it here:
https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#using-contexts-to-create-cache-keys

FredrikAugust (approved):

PR #33, opened 2021-02-18
Update component folder structure

### LINKED ISSUE #26: Create folder structure for future development
### Description
Implement an atomic design pattern, and make sure the rest of @IT2901-SINTEF01/frontend understand what this entails.

Some key concepts;
- Organisms: page structure
- Molecules: **responsible for data retrieval** and passing data down to atoms
- Atoms: **entirely self-contained** components. Should not require a parent* GraphQL provider. The demo by @emilom is a perfect example of this, where you could enter data in the "storybook configuration field" (hope you understand what I mean). This data should in practise be passed down by a molecule which uses a `useQuery` to retrieve from the GQL endpoint.

parent* = parent, grandparent, great-grandparent, &c

### Subtasks
- [x] Create folders and use `.keep` files if you need to push empty folders
- [x] Document component structure on wiki
- [x] Explain concept to rest of team

### Related sources
- https://bradfrost.com/blog/post/atomic-web-design/. The pitch above is a slightly simplified version of this, as we probably don't need templates and pages.

blauks:
Closes #26


FredrikAugust (approved):
Perfect! Wiki can be found [here](https://github.com/IT2901-SINTEF01/frontend/wiki/Folder-Structure).

Issue #30, opened 2021-02-17
Create dashboard organism


FredrikAugust:
### Description
Create a dashboard component which supports multiple child components (visualisations)

### Subtasks
- [x] Create `Dashboard` organism
- [x] Create sensible component which supports displaying multiple visualisation (might be useful to take a look at React children and `props.children`)
- [x] Decide how you should determine what data should be passed to what component. Do you wish to create very minimal wrapper components (molecules) which simply feed data into the respective visualisations (atoms)? I would recommend creating some examples and evaluating the different approaches. Feel free to send me a message if you want input or have any questions. **The decision should be documented in the wiki.**
- [x] Set a standard for how data transformation between raw data source and visualisation component input should be made. Perhaps create a generic type for this which ties into a prop to the visualisation component?

E.g.
```typescript
const rawMetApiDataToBarChar: (rawData: MetApiData) => BarChartProps['data'] = (rawData) => {
  // ...
}
```

### Related sources
- Depends on #26
- https://reactjs.org/docs/composition-vs-inheritance.html#containment

PR #45, opened 2021-02-24
Make dashboard items

blauks:
Part of #30

I have added a new dynamic Dashboard item, where two different views are available. In which one of them can have a graph and a description, while the other can only have a graph. The component works at any size.


FredrikAugust:
Please provide an explanation of the PR in the PR description.

src/dataTypes/DashboardItems/DashboardItemSmallType.tsx
```diff
@@ -0,0 +1,9 @@
+import { ReactNode } from 'react';
+
+export type DashboardItemSmallType = {
+    title: string;
+    titleSize: 100 | 200 | 300 | 400 | 500 | 600 | 700 | 800 | 900;
+    width: number | string;
+    height: number | string;
+    children: ReactNode;
+};
```
FredrikAugust: A few things worth mentioning here, so this will apply to more than just the lines mentioned;

- Files that do not contain Jsx content (e.g. `<Component>`) should have a `.ts` ending to clearly separate the two
- The folder name `dataTypes` could be simplified to `types`, or update to match the `mockdata` directory (all lowercase vs camelCase`
- Do you need a folder for the `DashboardItems`? There are only two very short files under it, so they could be merged into one file under (e.g.) `types/DashboardItem.ts`
- Types should not be suffixed by `Type` as this is implicit
- There is quite some overlap between the two types, and `DashboardItemSmallType` is a strict subset of `...LargeType`, so we can use some TS magic here to make the code DRYer (DRY=Do not Repeat Yourself):
```ts
export type DashboardItemSmall = {
  title: string;
  titleSize: 100 | 200 | 300 | 400 | 500 | 600 | 700 | 800 | 900;
  width: number | string;
  height: number | string;
}

export type DashboardItemLarge = DashboardItemSmall & {
  paragraph: string;
  textSize: 300 | 400 | 500;
}
```
Here you might find `interfaces` more intuitive, as they are written as;
```ts
interface IDashboardItemSmall {
  // ...
}

interface IDashboardItemLarge extends IDashboardItemSmall {
 // extra properties...
}
```
- Don't include children in your props. The definition of children here also doesn't encompass all acceptable React.children values. I don't know if this is intentional or not, but it might be worth noting. When you use the format of `React.FC<Props>`, `children` is automatically included from the generic `FC` (Functional Component) type.
- Generally speaking, if you wish to pass _one_ item of a type as a prop, it's better to use _a_ prop with that type instead of exploding the entire item into the props. Better demonstrated with an example;
```tsx
const Test: React.FC<{ name: string, age: number, pet: Animal }> = ({ name, age, pet }) => {
  // ...
}

// versus the more compact

const Test: React.FC<{ person: Person }> = ({ person }) => {
  // refer to person.name, person.age, &c
}
```
The benefits of this show when passing in the props;

```tsx
<Test name={person.name} age={person.age} pet={person.pet} />

// vs

<Test person={person} />
```

Consider [the line in this PR](https://github.com/IT2901-SINTEF01/frontend/pull/45/files#diff-d76616b8eab54f7d30f20cd8ccd132b403adf3a97e645f5c7b3c0e05d29b8c9aR9);
```tsx
const Template: Story<DashboardItemLargeType> = (args) => <DashboardItemLarge {...args} />;
```
What happens if there isn't a one-to-one relationship between the generic type `Story<DashoardItemLargeType>` and the expected props for `DashboardItemLarge`?

I would argue it's more canonical with;

```tsx
const Template: Story<ItemLarge> = (args) => (
  <DashboardItemLarge item={args.item} /> // or create item if necessary
);
```

FredrikAugust (changes requested):
Great work! Added some thoughts. Do with them as you please!:)

The required changes are mainly the proptype changes.

src/dataTypes/DashboardItems/DashboardItemSmallType.tsx
FredrikAugust: Regarding the last lines;

You might need to look into `PropsWithChildren` (or something along those lines) for the generic `Story` type.

blauks: I will look into this!

src/types/DashboardItem.ts
```diff
@@ -1,11 +1,11 @@
-import { ReactNode } from 'react';
-
-export type DashboardItemLargeType = {
+export type DashboardItemSmall = {
```
FredrikAugust: Consider moving Props into the files in which the components are created, and name them accordingly;

`DashbaordItemSmallProps`

src/stories/DashboardItemSmall.stories.tsx
```diff
@@ -28,10 +21,23 @@ Primary.args = {
...
+    </DashboardItemSmall>
+);
+
+export const Small = Template.bind({});
+Small.args = {
+    title: 'This is a title',
+    width: 500,
+    height: 300,
+    titleSize: 500,
 };
```
FredrikAugust: You can move this into the export default

src/types/DashboardItem.ts
blauks: In this case, is it not better to have both in the same file? Since one of them depends on the other.

FredrikAugust: I believe it's fine to import the type. You can use a type-only import to ensure the coupling between the files is loose like this;

```typescript
export type DashboardItemLargeProps = import("./ComponentSmall"). DashboardItemSmall & {

}
```

or if you find it cleaner, import the small-type first on a separate line with `type BlaBlaSmall = import(...);`

src/components/molecules/DashboardItem.tsx
```diff
@@ -0,0 +1,62 @@
...
+    titleSize,
+    textSize,
+    children,
+}) => {
+    return (
+        <Card width={width} height={height} display="flex" flexDirection="column" border="default" elevation={1}>
+            <Heading size={titleSize} marginTop={25} marginLeft={25}>
+                {title}
+            </Heading>
+            {paragraph && textSize ? (
```
FredrikAugust: Seems a bit strange to include the presence of the `textSize` property in the conditional. Could it perhaps have a default value of `300 | 400 | 500` and thus only include the `paragraph` in the conditional?

src/stories/DashboardItem.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+    paragraph:
+        'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.',
+    textSize: 300,
+};
+Large.argTypes = {
+    width: { control: { type: 'range', min: 100, max: 1200, step: 10 } },
+    height: { control: { type: 'range', min: 100, max: 1800, step: 10 } },
+    textSize: { control: { type: 'radio', options: [300, 400, 500] } },
+    titleSize: { control: { type: 'select', options: [100, 200, 300, 400, 500, 600, 700, 800, 900] } },
+};
```
FredrikAugust, line 54: Sooo much cleaner than the original code! 🚀

FredrikAugust (approved):

FredrikAugust (dismissed):

src/components/molecules/DashboardItem.tsx
blauks: True! I can add a default value instead

src/components/molecules/DashboardItem.tsx
```diff
@@ -35,7 +35,7 @@ const DashboardItem: React.FC<DashboardItemProps> = ({
                     alignSelf="center"
                     marginTop={10}
                 >
-                    <Paragraph width="15%" height="100%" size={textSize}>
+                    <Paragraph width="15%" height="100%" size={textSize || 300}>
```
FredrikAugust: You can set this in the prop list instead to make it _a proper_ default value.

FredrikAugust: `textSize = 300,`

FredrikAugust (dismissed):

src/components/molecules/DashboardItem.tsx
blauks: On it:)

FredrikAugust (approved):
LGTM!
In the future, it's better to either replace the code or revert it, rather than amend (?) and force-push. Force-pushing is generally something you should avoid as you replace the history.

blauks:
Okie dokie!

PR #46, opened 2021-02-24
Make interactive Dashboard 🕺 

emilom:
Part of #30 

Made dashboard with react-grid-layout🍝


FredrikAugust:
Note that this PR appears to depend on #45.
I believe we should aim to resolve the issues in that PR, rebase that onto master, and then into this branch again before merging this PR.
I will review this PR shortly, but will not comment on the points discussed in #45, so refer to that for additional comments.

sider.yml
```diff
@@ -23,7 +23,7 @@ linter:
             - cypress
         #       - bar.rb
         locale: UK
-#     ignore: center,behavior
+        ignore: center,behavior,color
```
FredrikAugust, line 26: Generally, try to avoid adding extra exceptions that aren't needed. Ref. "behavior".

src/components/molecules/DashboardItem.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import { Card } from 'evergreen-ui';
+
+const DashboardItem: React.FC = () => {
+    console.log('title');
```
FredrikAugust, line 5: Don't use `console.log` in React components

src/components/molecules/DashboardItem.tsx
```diff
@@ -0,0 +1,9 @@
...
+    return <Card width={'100%'} height={'100%'} border="default" elevation={1}></Card>;
```
FredrikAugust, line 6: Don't need to evaluate the strings as they are _simply strings_.
```suggestion
    return <Card width="100%" height="100%" border="default" elevation={1} />;
```

This is because statements wrapped in `{}` are evaluated as JS expressions.

src/components/molecules/DashboardItemSmall.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { DashboardItemSmallType } from '../../dataTypes/DashboardItems/DashboardItemSmallType';
+import { Card, Pane, Heading } from 'evergreen-ui';
+
+const DashboardItemSmall: React.FC<DashboardItemSmallType> = ({ title, children, width, height, titleSize }) => {
```
FredrikAugust, line 5: See #45.

src/components/organisms/Dashboard.tsx
```diff
@@ -0,0 +1,148 @@
...
+import ReactGridLayout from 'react-grid-layout';
+import { ParentSize } from '@visx/responsive';
+
+import { Pane, MinimizeIcon, MaximizeIcon, IconButton } from 'evergreen-ui';
+
+import DashboardItem from '../molecules/DashboardItem';
+import { DashboardItemSize, DashboardItemInfo, DashboardProps } from '../../dataTypes/Dashboard/DashboardTypes';
+
+import 'react-grid-layout/css/styles.css';
+import 'react-resizable/css/styles.css';
```
FredrikAugust, line 11: Include this in `index.tsx` to ensure they're only loaded once.

src/components/organisms/Dashboard.tsx
```diff
@@ -0,0 +1,148 @@
...
+                    x += getItemWidth(item.size);
+                }
+            }
+        });
+        setLayout(gridLayout);
+    }, [items]);
+
+    return (
+        <Pane height={'100%'} width={'80%'} marginLeft={'10%'}>
+            {console.log('layout change')}
```
FredrikAugust:
```suggestion
```

src/components/organisms/Dashboard.tsx
```diff
@@ -0,0 +1,148 @@
...
+                    });
+                    x += getItemWidth(item.size);
+                }
+            }
+        });
+        setLayout(gridLayout);
+    }, [items]);
+
+    return (
+        <Pane height={'100%'} width={'80%'} marginLeft={'10%'}>
```
FredrikAugust: Prev comment.

src/components/organisms/Dashboard.tsx
```diff
@@ -0,0 +1,148 @@
...
+type Layout = {
+    i: string;
+    x: number;
+    y: number;
+    w: number;
+    h: number;
+};
+
+const Dashboard: React.FC<DashboardProps> = ({ items }) => {
+    const [layout, setLayout] = useState<Layout[]>();
```
FredrikAugust:
```suggestion
    const [layout, setLayout] = useState<Layout[]>([]);
```
Default to empty layout instead of `undefined`/`null`.

src/components/organisms/Dashboard.tsx
```diff
@@ -0,0 +1,148 @@
...
+                        className="layout"
+                        layout={layout}
+                        cols={6}
+                        rowHeight={visWidth / 6}
+                        width={visWidth}
+                        verticalCompact={true}
+                        useCSSTransforms={true}
+                        isResizable={false}
+                    >
+                        {layout &&
```
FredrikAugust: Remove conditional rendering with default `[]` in `setState`. No longer necessary.

src/dataTypes/Dashboard/DashboardTypes.ts
```diff
@@ -0,0 +1,14 @@
+export type DashboardItemInfo = {
+    size: DashboardItemSize;
+    withText: boolean;
+};
+
+export enum DashboardItemSize {
+    small,
+    medium,
+    large,
```
FredrikAugust:
```suggestion
    SMALL = 2,
    MEDIUM = 3,
    LARGE = 4,
```

Default is all-caps for enums.
Set the values for the enum values here instead of retrieving them later.

src/dataTypes/Dashboard/DashboardTypes.ts
```diff
@@ -0,0 +1,14 @@
...
+}
+
+export type DashboardProps = {
+    items: DashboardItemInfo[];
+};
```
FredrikAugust, line 14: Props should be in the same file as the component.

src/components/organisms/Dashboard.tsx
```diff
@@ -0,0 +1,148 @@
...
+import { Pane, MinimizeIcon, MaximizeIcon, IconButton } from 'evergreen-ui';
+
+import DashboardItem from '../molecules/DashboardItem';
+import { DashboardItemSize, DashboardItemInfo, DashboardProps } from '../../dataTypes/Dashboard/DashboardTypes';
+
+import 'react-grid-layout/css/styles.css';
+import 'react-resizable/css/styles.css';
+
+type Layout = {
+    i: string;
```
FredrikAugust: It seems like there are more conversions from int to string than reverse, so you could type this as an int.

And with that replace `Layout` with `Layout = Record<'i' | 'x' | 'y' | 'w' | 'h', number>;`.

src/index.css
```diff
@@ -0,0 +1,9 @@
+html {
+    height: 100%;
+}
+
+body {
+    margin: 0;
```
FredrikAugust, line 6: Consider a CSS reset

src/index.css
```diff
@@ -0,0 +1,9 @@
+html {
+    height: 100%;
```
FredrikAugust, line 2: Perhaps use `100vh` if you wish to limit to the full screen height.

src/stories/DashboardItemSmall.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+                    yLabel="Price"
+                />
+            )}
+        </ParentSize>
+    ),
+};
+
+export default {
+    title: 'Dashboard/Dashboard Item Small',
+    component: DashboardItemSmall,
```
FredrikAugust, line 36: Use `Primary` as the component.

Also you don't need to use the `.bind` as far as I can see. 

Take a look at the rewritten `Dashboard.stories` file I pushed.

FredrikAugust (changes requested):
Pushing up some changes that we can take a look at tomorrow

FredrikAugust:
Due to issues with performance, I've temporarily removed the min/max feature on the cards to simplify the currently semi-functional code.

There's still quite a lot of unnecessary renders being performed by the grid system, which you can see with the profiling features in the React browser toolkit, or by the `console.log` statements in your console.

Issue #32, opened 2021-02-17
Design for loading and error


FredrikAugust:
### Description
The datasets might take some time to load, and may even produce errors. We should have visual components representing these outcomes.

### Subtasks
- [x] Create loading and error designs for datasets, visualisations and other elements which require loading of external resources

FredrikAugust:
@thomran any updates on this?

thomran:
Should be completed later today! (02.03.2021)

FredrikAugust:
Superb!

FredrikAugust:
@thomran could you also create a component for when no data is retrieved from the backend?

FredrikAugust:
I.e. missing data.

thomran:
On it!

FredrikAugust:
Perfect!

thomran:
Done! Will fill out wiki with new content tomorrow. 

Will also need to discuss some aspects of the design in tomorrow's meeting.

PR #49, opened 2021-03-02
Create dashboard and DataWrapper 

### LINKED ISSUE #30: Create dashboard organism

### LINKED ISSUE #51: Set up a mocked provider for use in storybook
### Description
To provide some more interesting data to the components in the storybook (as well as avoiding having to store large JSON files everywhere) we can use a mocked provider from apollo.

### Subtasks
- [x] Set up a [decorator](https://storybook.js.org/docs/react/writing-stories/decorators) which provides the mocked provider
- [x] Mock the api.met.no request

emilom:
Closes #30
Closes #51


src/index.css
```diff
@@ -0,0 +1,8 @@
+html {
+    height: 100vh;
+    width: 100%;
+}
+
+body {
+    margin: 0;
+}
```
blauks:
```suggestion
}

```

emilom:
This is mostly done,  storybook is coming. Had some problems with apollo provider in storybook. @blauks and I will finish this today

src/utils/dashboardList.ts
```diff
@@ -0,0 +1,11 @@
...
+
+/*
+export const generateRandomList = (size: number): DashboardItemInfo[] =>
+    new Array(size).fill(undefined).map(() => ({
+        size: [DashboardItemSize.SMALL, DashboardItemSize.MEDIUM, DashboardItemSize.LARGE][
+            Math.floor(Math.random() * 3)
+        ],
+        withText: false,
+    }));
+*/
```
FredrikAugust: Try to avoid commented-out code in PRs.

src/types/dashboard.ts
```diff
@@ -0,0 +1,15 @@
+//Can be expanded with visualisation method if needed.
+//Hva vi må legge til. Text. Title. Visualisation method. Query. //Source.
```
FredrikAugust: :gb:

src/stories/Dashboard.stories.tsx
```diff
@@ -0,0 +1,11 @@
...
+import { Meta } from '@storybook/react/types-6-0';
+import { Story } from '@storybook/react';
+import Dashboard from '../components/organisms/Dashboard';
+
+export const Primary: React.FC = () => <Dashboard />;
+
+export default {
+    title: 'Component/Dashboard',
+    component: Dashboard,
+} as Meta;
```
FredrikAugust, line 10: Component isn't Dashboard, it's Primary.
@blauks might be able to give you some pointers here.

src/queries/metApi.ts
```diff
@@ -0,0 +1,24 @@
+import { gql } from '@apollo/client';
+
+//Format: What where
```
FredrikAugust: What does this mean?

src/index.tsx
```diff
@@ -1,6 +1,23 @@
...
+import { ApolloProvider } from '@apollo/client';
+import { cache } from './cache/cache';
 
-import Filler from './components/atoms/Filler';
+import Dashboard from './components/organisms/Dashboard';
+import './index.css';
 
-render(<Filler />, document.getElementById('root'));
+const client = new ApolloClient({
+    uri: '/graphql',
```
FredrikAugust: This should be read from an environment variable

src/index.tsx
```diff
@@ -1,6 +1,23 @@
...
+    cache,
+});
+
+const App: React.FC = () => {
+    return (
+        <ApolloProvider client={client}>
+            <Dashboard />
+        </ApolloProvider>
+    );
+};
```
FredrikAugust: Move this to its own file.

src/index.tsx
```diff
@@ -1,6 +1,23 @@
 import React from 'react';
 import { render } from 'react-dom';
+import { ApolloClient } from '@apollo/client';
+import { ApolloProvider } from '@apollo/client';
+import { cache } from './cache/cache';
```
FredrikAugust: Do you need a folder for cache with a file called cache?
If it's the "root" cache it's common to name it `index.ts`

src/types/dashboard.ts
blauks: The norwegian comment can be removed
```suggestion
// Can be expanded with visualisation method if needed.
```

src/index.css
FredrikAugust:
```suggestion
}

```

As preferred by Dr. @blauks

src/components/organisms/Dashboard.tsx
```diff
@@ -0,0 +1,70 @@
...
+                    <option value="bar">Mitt andre dashboard</option>
+                </Select>
+                <Text position="absolute" bottom="25px" left="220px" color="#1070CA" cursor="pointer">
+                    + Legg til dashboard
+                </Text>
+            </Pane>
+            <Pane
+                width="80%"
+                display="grid"
+                gridTemplateColumns="1fr 1fr 1fr 1fr 1fr 1fr" //Smaller screen size should have 1f 1fr 1f
```
FredrikAugust, line 45: You can also use `repeat(n, 1fr)` where `n` is the number of columns.

FredrikAugust: At least you can do that in CSS, so I believe it should work here as well.

src/components/organisms/Dashboard.tsx
```diff
@@ -0,0 +1,70 @@
...
+import DashboardItem from '../molecules/DashboardItem';
+import { useReactiveVar } from '@apollo/client';
+import { dashboardItemsVar } from '../../cache/cache';
+import DataWrapper from '../molecules/DataWrapper';
+import ThresholdChart from '../atoms/ThresholdChart';
+import { ParentSize } from '@visx/responsive';
+
+const Dashboard: React.FC = () => {
+    //Apollo local state
+    const dashboardItems = useReactiveVar(dashboardItemsVar);
```
FredrikAugust, line 14: Due to this, you should wrap your storybook template in a provider as well.

@blauks knows more about this (regarding templates and exports).

src/components/molecules/DataWrapper.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { TimeEntry } from '../atoms/ThresholdChart';
+import { formatMetData } from '../../utils/formatMetData';
+
+export type DataWrapperProps = {
+    query: DocumentNode;
+    children: (args: { data: TimeEntry[] | undefined; loading: boolean; error: ApolloError | undefined }) => ReactNode;
+};
+
+const DataWrapper: React.FC<DataWrapperProps> = ({ children, query }) => {
+    const [formattedData, setFormattedData] = useState<TimeEntry[] | undefined>(undefined);
```
FredrikAugust:
```suggestion
    const [formattedData, setFormattedData] = useState<TimeEntry[]>([]);
```

FredrikAugust: Default to empty formattedData instead might be easier to type. Just a suggestion.

If not, I believe you can omit `| undefinned` and the `undefined` in the argument list, and it will default to `null`, which I believe is more semantically correct here.

src/components/molecules/DataWrapper.tsx
```diff
@@ -0,0 +1,22 @@
...
+    const { loading, error, data } = useQuery(query);
+
+    useEffect(() => {
+        if (data) setFormattedData(formatMetData(data));
+    }, [data]);
+
+    return <>{children({ data: formattedData, loading, error })}</>;
```
FredrikAugust: What about handling `loading` and `error` in this level instead of passing it down?

That way you would remove one area of responsibility from the child.

FredrikAugust: Since `data` will only change once, you could structure it like this (presuming you handle `loading`/`error` in this component as well)

```tsx
if (loading) {
  return <Loading />;
}

if (error) {
  return <Error error={error} />;
}

return children(data);
```

This would make typing the child a lot easier.

src/components/molecules/DataWrapper.tsx
```diff
@@ -0,0 +1,22 @@
...
+    query: DocumentNode;
+    children: (args: { data: TimeEntry[] | undefined; loading: boolean; error: ApolloError | undefined }) => ReactNode;
+};
+
+const DataWrapper: React.FC<DataWrapperProps> = ({ children, query }) => {
+    const [formattedData, setFormattedData] = useState<TimeEntry[] | undefined>(undefined);
+    const { loading, error, data } = useQuery(query);
+
+    useEffect(() => {
+        if (data) setFormattedData(formatMetData(data));
```
FredrikAugust: Here you limit this component to only working with `metData`. Could you pass this mapping function in from props?

src/components/atoms/AddComponentButton.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { Card, Text } from 'evergreen-ui';
+import { dashboardItemsVar } from '../../cache/cache';
+import { WEATHER_MET_API } from '../../queries/metApi';
+
+const AddComponentButton: React.FC = () => {
+    const addDashboardItem = () => {
+        dashboardItemsVar([...dashboardItemsVar(), { size: 3, withText: false, query: WEATHER_MET_API }]);
+    };
```
FredrikAugust: This causes the storybook to crash when you try to click the button. You should either mock it in some way or provide a provider in the storybook example.

src/index.css
```diff
@@ -5,4 +5,4 @@ html {
 
 body {
     margin: 0;
-}
\ No newline at end of file
+}
```
FredrikAugust, line 8: funny man with the magic words 🧙🏻 you just outperformed me

src/components/organisms/Dashboard.tsx
emilom: I tried, but it didn’t work with evergreen. It might be possible but just passing the string  "repeat(6, 1fr)" did not work.

src/components/molecules/DataWrapper.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { ReactNode, useEffect, useState } from 'react';
+import { ApolloError, DocumentNode, useQuery } from '@apollo/client';
+import { TimeEntry } from '../atoms/ThresholdChart';
+import { formatMetData } from '../../utils/formatMetData';
+
+export type DataWrapperProps = {
+    query: DocumentNode;
+    children: (args: { data: TimeEntry[] | undefined; loading: boolean; error: ApolloError | undefined }) => ReactNode;
```
FredrikAugust: This limits the DataWrapper to only work with the `ThresholdChart` (only accepting `data: TimeEntry[]`).

I believe this would be a fitting time to introduce ✨ _generics_ ✨ 

In essence, you wish to create to create a `DataWrapper` which takes
- a `query` which resolves to a type `T` (for use in `useQuery<T>(query)`)
- a mapping function mapping from type `T -> U`, where `U` is the input to the child function
_What does this mean?_ Albeit a slightly functional (and specifically Haskell) explanation, it essentially means you have to provide a function (in the props) which takes the **query data type** (`T`, e.g. `MetApiRawData`) and maps it to **some arbitrary output type** (`U`). In your example, you have provided this, only hard-coded. 

`formatMetData` takes the raw data type `T` as input, and produces some output type `U` which is further used by the children (in this example, `TimeEntry[]`). 

Thus we can say that `formatMetData` is a mapping from `T -> U`.

If you further eliminate most data from the child function, so you're left with `children: (data: U) => ReactNode`, this can be typed quite neatly, and you have provided the following;

- a **generic** (:sparkles: :mage:) component which retrieves an arbitrary GraphQL data source and provides it to a child visualisation. Given that you handle the error/loading states in this component, the child component (read: not the child function) can have a prop type like `props: { data: TimeEntry[] }` (note that it's not nullable) as you guarantee that it will be provided with actual data.

The usage would thus look like;

```tsx
<DataWrapper<MetAPIRawData, TimeEntry[]>
  mapping={(entry) => ({ x: entry.x, y: entry.y })}
  query={I_LOVE_THE_MONKEY_HEAD_QUERY}
>
  {(data) => <ThresholdChart data={data} />}
</DataWrapper>
```

src/queries/metApi.ts
emilom: Describing the naming format for queries. What = WEATHER and where = MET_API.

src/components/organisms/Dashboard.tsx
```diff
@@ -0,0 +1,70 @@
...
+                                                    {loading && <Text>Loading...</Text>}
+                                                    {error && <Text>{error.message}</Text>}
+                                                    {data && (
+                                                        <ThresholdChart
+                                                            data={data}
+                                                            height={parent.height}
+                                                            width={parent.width}
+                                                        />
+                                                    )}
+                                                </>
```
FredrikAugust: This error handling doesn't belong in the `organism`. It only concerns the data, which is handled by `molecule`.

src/components/organisms/Dashboard.tsx
```diff
@@ -0,0 +1,70 @@
...
+            >
+                {dashboardItems.map((item, i) => {
+                    return (
+                        <Pane key={i} width="100%" height="100%" gridColumn={'span ' + item.size}>
+                            <DashboardItem title="placeholder" height="100%" width="100%" titleSize={100}>
+                                <ParentSize>
+                                    {(parent) => (
+                                        <DataWrapper query={item.query}>
+                                            {({ data, loading, error }) => (
+                                                <>
```
FredrikAugust: Loading, error and data are exclusive states -- i.e. they cannot co-exist, which I believe can be conveyed in a cleaner format.

FredrikAugust: Read molecule comments

src/components/organisms/Dashboard.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { ParentSize } from '@visx/responsive';
+
+const Dashboard: React.FC = () => {
+    //Apollo local state
+    const dashboardItems = useReactiveVar(dashboardItemsVar);
+
+    return (
+        <Pane width="100%" height="100%" display="flex" flexDirection="column" alignItems="center" paddingBottom="20px">
+            <Pane width="80%" height="100px" position="relative">
+                <Select position="absolute" bottom="20px" width="200px">
```
FredrikAugust: Is it possible to avoid using pixels?

`rem`s are generally better if you don't need very specific control over an elements size.

src/components/atoms/AddComponentButton.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+    return (
+        <Card
+            cursor="pointer"
+            width="100%"
+            height="100%"
+            gridColumn={'span 2'}
```
FredrikAugust:
```suggestion
            gridColumn="span 2"
```

src/components/atoms/AddComponentButton.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { Card, Text } from 'evergreen-ui';
+import { dashboardItemsVar } from '../../cache/cache';
+import { WEATHER_MET_API } from '../../queries/metApi';
+
+const AddComponentButton: React.FC = () => {
+    const addDashboardItem = () => {
+        dashboardItemsVar([...dashboardItemsVar(), { size: 3, withText: false, query: WEATHER_MET_API }]);
```
FredrikAugust: Ref. discussion with Jonathan, do you need `withText`? Or can you infer it from the content of the data/components?

src/cache/cache.ts
```diff
@@ -0,0 +1,18 @@
...
+            fields: {
+                dashboardItems: {
+                    read() {
+                        return dashboardItemsVar();
+                    },
+                },
+            },
+        },
+    },
+});
```
FredrikAugust: Cool! Never seen it in use before.

FredrikAugust (changes requested):

src/components/molecules/DataWrapper.tsx
emilom: That could work. Were thinking about using a switch based on the query to know which format that should be used. Not sure what the best approach is.

FredrikAugust:
UI Tests denied as it's broken. Ref other comments.

src/components/molecules/DataWrapper.tsx
FredrikAugust: Further information:

If you don't feel comfortable with generics:
- https://www.typescriptlang.org/docs/handbook/generics.html

- Google "react generic components"
- https://wanago.io/2020/03/09/functional-react-components-with-generic-props-in-typescript/

src/components/molecules/DataWrapper.tsx
FredrikAugust: Generics! They are a perfect fit for this (at least I think so):)

src/queries/metApi.ts
FredrikAugust: OK. Could it be expressed in a clearer way?

E.g.

```tsx
// Naming convention:
// <RESOURCE>_<SOURCE>
```

src/queries/metApi.ts
```diff
@@ -0,0 +1,24 @@
...
+                            details {
+                                airTemperature
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+`;
```
FredrikAugust, line 25: I'd also recommend typing the result in this file, as @blauks witnessed in ekskom.

So you export the GraphQL query as well as 

```ts
export type WeatherMetApi = {
  // type type type
}
```

FredrikAugust:
It also appears that this PR broke the docker registry configuration.
It would be great if you could take a look at this.
If you cannot find any reason for why this is happening, you can replace

```diff
- # RUN yarn build
- RUN npx react-scripts build
+ RUN yarn build
```

src/index.tsx
FredrikAugust: And thus also document the environment variable in README+wiki.

src/utils/dashboardList.ts
FredrikAugust: Also check commit comments for more information

src/stories/Dashboard.stories.tsx
blauks: Hey, @blauks here! According to the docs you are supposed to use the imported component, and not a story. So this can be resolved @emilom

src/components/atoms/AddComponentButton.tsx
emilom: changed the type to `description?: string`

src/queries/metApi.ts
FredrikAugust: By the way, the intention behind these comments is not to tell you what to do, so don't feel obligated to incorporate them. It's rather to express my thoughts and reactions.

If you feel the suggestions are in any way suboptimal, don't hesitate to provide counter-arguments or initiate a debate (med vennlig hilsen Fredrik Solvang).

src/stories/Dashboard.stories.tsx
FredrikAugust: Noted! Good catch. 

Close this when you see it @emilom.

FredrikAugust: I suppose it's there to tell you which component the story is describing?

FredrikAugust:
By the way, it would be nice if you could write some more specific commit messages. This makes it a bit easier to keep track of what's happening during the review.

E.g. `Fix dashboard type` could be `Use optional paragraph instead of withText in DashboardItemInfo`

It's worth spending a short minute coming up with a good commit message.

src/components/molecules/DataWrapper.tsx
emilom: How do you return just `children(data)`. I can’t get it to work without wrapping it in `<>{children(data)}</>`

FredrikAugust: I don't know, I just made something up.

You can try
- looking at the stack trace. what went wrong?
- what type did it produce, and what does it expect?
- googling the issue
- taking a look at the typing in other examples

FredrikAugust: For reference, you're wrapping the child function invocation in a fragment. To understand why it doesn't work without, it might be helpful to understand what this does. 

It might also be helpful to research these kinds of issues before asking in the code review. If you're still stuck after some time, you can in #help-programming in Slack. I'd appreciate it if we could (for the most part) try to keep code reviews to discussions rather than programming questions.

src/components/organisms/Dashboard.tsx
```diff
@@ -47,19 +47,15 @@ const Dashboard: React.FC = () => {
...
-                                                        <ThresholdChart
-                                                            data={data}
-                                                            height={parent.height}
-                                                            width={parent.width}
-                                                        />
-                                                    )}
-                                                </>
-                                            )}
+                                            {({ data }) =>
+                                                data.length !== 0 && (
```
FredrikAugust: Is this a necessary check?

In the case where it's zero you are we are now rendering _nothing_ instead of an empty chart.

src/components/molecules/DataWrapper.tsx
```diff
@@ -1,22 +1,29 @@
...
         if (data) setFormattedData(formatMetData(data));
     }, [data]);
 
-    return <>{children({ data: formattedData, loading, error })}</>;
+    if (loading) {
+        return <p>Loading...</p>; //Change to loading atom
+    }
+    if (error) {
+        return <p>Error :/</p>; //Change to error atom
+    }
```
FredrikAugust: Create issues for this if it's not going to be addressed in this PR.

FredrikAugust: Ties up to #32 which is scheduled for completion today.

src/components/molecules/DataWrapper.tsx
```diff
@@ -1,22 +1,29 @@
...
     useEffect(() => {
         if (data) setFormattedData(formatMetData(data));
     }, [data]);
 
-    return <>{children({ data: formattedData, loading, error })}</>;
+    if (loading) {
+        return <p>Loading...</p>; //Change to loading atom
+    }
+    if (error) {
+        return <p>Error :/</p>; //Change to error atom
```
FredrikAugust: Might be helpful to render `error.message` for the time being while a proper atom is being constructed.

FredrikAugust (dismissed):
I'm unable to comment on the line, but it's not recommended to use indexes as keys.

> We don’t recommend using indexes for keys if the order of items may change. This can negatively impact performance and may cause issues with component state. Check out Robin Pokorny’s article for an in-depth explanation on the negative impacts of using an index as a key. If you choose not to assign an explicit key to list items then React will default to using indexes as keys.

https://reactjs.org/docs/lists-and-keys.html#keys

src/components/organisms/Dashboard.tsx
FredrikAugust: You could also perform a check in the `DataWrapper` component to see if data is a falsy value.

emilom: ThresholdChart does not work with an empty array, so it will throw an error

FredrikAugust: OK. Perhaps it would be wise to show an error if the data array is empty?

If you don't have capacity to fix this in this PR, you can create an issue for it.

emilom: Sounds good. I'll create an issue as the design for error messages and loading is due today.

emilom: new issues #50

src/components/molecules/DataWrapper.tsx
emilom: new issues #50

src/components/molecules/DataWrapper.tsx
FredrikAugust: `formatMetData` (or the equivalent function (mapping `T->U`)) can also be used in the `useQuery` call;

https://www.apollographql.com/docs/react/data/queries/#options

Look at `onCompleted`

FredrikAugust: Note that this isn't relevant if you remove the `setFormattedData` call, but it can be useful to note.

src/components/organisms/Dashboard.tsx
blauks: Probably!

src/components/organisms/Dashboard.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { formatMetData } from '../../utils/formatMetData';
+
+const Dashboard: React.FC = () => {
+    //Apollo local state
+    const dashboardItems = useReactiveVar(dashboardItemsVar);
+
+    return (
+        <Pane width="100%" height="100%" display="flex" flexDirection="column" alignItems="center" paddingBottom="20px">
+            <Pane width="80%" height="100px" position="relative">
+                <Select position="absolute" bottom="20px" width="200px">
```
FredrikAugust: Why do these two use relative/absolute positioning? 

Is it possible to position them using grid/flex instead?

emilom: should be possible 👍

emilom: just wanted to see what it would look like

FredrikAugust:
Alright.

If it’s just poking around to see how things react, try to avoid pushing it for review.

src/components/molecules/DataWrapper.tsx
blauks: Thanks for the good tips! I will definitely look more into generic types:)

src/components/organisms/Dashboard.tsx
blauks: Done

blauks:
> It also appears that this PR broke the docker registry configuration.
> It would be great if you could take a look at this.
> If you cannot find any reason for why this is happening, you can replace
> 
> ```diff
> - # RUN yarn build
> - RUN npx react-scripts build
> + RUN yarn build
> ```

This did not work. I will google some more

.dockerignore
```diff
@@ -1,6 +1,8 @@
 node_modules
 .eslint*
+.eslintrc.js
 .prettier*
+.prettierrc.js
```
FredrikAugust: The two `.js` files are already handled by the wildcard-ignores.

blauks:
Small error with threshold chart hehe, no need for UI review right now.

FredrikAugust (approved):

PR #52, opened 2021-03-05
Add loading and error from evergreen to DataWrapper

### LINKED ISSUE #50: Create atom for loading and error
### Description
We need custom atoms for loading and errors which can be displayed when fetching data from the backend. 

### Subtasks
- [x] Create atom for loading as designed in #32 
- [x] Create atom for error as designed in #32
- [x] Create atom for cases where GraphQL didn't fetch any data (as designed in #32)
- [x] Implement atoms in DataWrapper.tsx and Dashboard.tsx

emilom:
Closes #50 

DataWrapper will now handle loading and error messages from GraphQL. This will be displayed on the dashboard in DashboardItems.

### DIFF
--- a/src/components/atoms/ErrorMessage.tsx
+++ b/src/components/atoms/ErrorMessage.tsx
@@ -1,0 +1,29 @@
+import { InlineAlert } from 'evergreen-ui';
+import React from 'react';
+import { Text, Pane } from 'evergreen-ui';
+
+export type ErrorMessageProps = {
+    size?: 300 | 400 | 500 | 600;
+    message?: string;
+    moreInfo?: string;
+};
+
+const ErrorMessage: React.FC<ErrorMessageProps> = ({ message = 'Error :(', moreInfo, size = 400 }) => {
+    return (
+        <Pane
+            display="flex"
+            flexDirection="column"
+            width="100%"
+            height="100%"
+            justifyContent="center"
+            alignItems="center"
+        >
+            <InlineAlert intent="danger" size={size}>
+                {message}
+            </InlineAlert>
+            {moreInfo && <Text size={size}>{moreInfo}</Text>}
+        </Pane>
+    );
+};
+
+export default ErrorMessage;
--- a/src/components/atoms/ErrorMessage.tsx
+++ b/src/components/atoms/ErrorMessage.tsx
@@ -1,6 +1,5 @@
-import { InlineAlert } from 'evergreen-ui';
 import React from 'react';
-import { Text, Pane } from 'evergreen-ui';
+import { Text, Pane, InlineAlert } from 'evergreen-ui';
 
 export type ErrorMessageProps = {
     size?: 300 | 400 | 500 | 600;
--- a/src/components/molecules/DataWrapper.tsx
+++ b/src/components/molecules/DataWrapper.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { DocumentNode, useQuery } from '@apollo/client';
-import { InlineAlert, Pane, Spinner } from 'evergreen-ui';
+import Loading from '../atoms/Loading';
+import ErrorMessage from '../atoms/ErrorMessage';
 
 /**
  * @param T The raw output of the GraphQL query.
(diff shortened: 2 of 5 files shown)


blauks (commented):
Hey Emil, this LGTM to me! But I think it would be even better if you added a Storybook item for each of the different scenarios!🙂

emilom:
> Hey Emil, this LGTM to me! But I think it would be even better if you added a Storybook item for each of the different scenarios!🙂

Great idea! Thanks for the input, will take a look at storybook now 👊

src/components/molecules/DataWrapper.tsx
```diff
@@ -11,23 +12,44 @@ export type DataWrapperProps<T, U> = {
...
 };
 
+const CenterPane: React.FC = ({ children }) => {
+    return (
+        <Pane width="100%" height="100%" display="flex" justifyContent="center" alignItems="center">
+            {children}
+        </Pane>
+    );
+};
+
```
FredrikAugust: Perhaps move to atom if it's going to be reused other places?

src/components/molecules/DataWrapper.tsx
```diff
@@ -11,23 +12,44 @@ export type DataWrapperProps<T, U> = {
...
     }
 
     if (error) {
-        return <p>Error: {error.message}</p>;
+        return (
+            <CenterPane>
+                <InlineAlert intent="danger">Vi finner ikke dataen du leter etter.</InlineAlert>
+            </CenterPane>
+        );
     }
```
FredrikAugust: Få med hva error-en er. Dette kan du finne i `error.message`.

FredrikAugust: Få med hva error-en er. Dette kan du finne i `error.message`. 

You should already know this.

FredrikAugust: Kanskje få med at et problem oppstod under henting av data. Typ "Det oppsto en feil under henting av data".

FredrikAugust (changes requested):
Will the toasts be implemented in another PR? If so we should create a work package for that.

src/components/atoms/ErrorMessage.tsx
```diff
@@ -0,0 +1,29 @@
+import { InlineAlert } from 'evergreen-ui';
+import React from 'react';
+import { Text, Pane } from 'evergreen-ui';
```
blauks:
```suggestion
import React from 'react';
import { Text, Pane, InlineAlert } from 'evergreen-ui';
```

src/components/atoms/ErrorMessage.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export type ErrorMessageProps = {
+    size?: 300 | 400 | 500 | 600;
+    message?: string;
+    moreInfo?: string;
+};
+
+const ErrorMessage: React.FC<ErrorMessageProps> = ({ message = 'Error :(', moreInfo, size = 400 }) => {
```
FredrikAugust, line 10: A nifty trick is to replace the 

```
{
  return (
```

with just `(` (meaning implicit return).

You don't need to change this, but it is a helpful tool.

src/stories/ErrorMessage.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+const sizeControl = { control: { type: 'radio', options: [300, 400, 500, 600] } };
+
+export const Primary = Template.bind({});
+Primary.args = { size: 400, message: 'Error message' };
+Primary.argTypes = {
+    size: sizeControl,
+};
+
+export const Secondary = Template.bind({});
+
```
blauks, line 24:
```suggestion
```

blauks: Just to keep everything nice and equal :)

emilom:
> Will the toasts be implemented in another PR? If so we should create a work package for that.

No need for toast now, I'm not hungry 😋 
We might use it later.

FredrikAugust (dismissed):

blauks:
> > Will the toasts be implemented in another PR? If so we should create a work package for that.
> 
> No need for toast now, I'm not hungry 😋
> We might use it later.

Maybe for dinner 😁

FredrikAugust (approved):

PR #62, opened 2021-03-15
Added data source selection page and data result item

### LINKED ISSUE #53: Create data source selection page
### Description
Good morning!

Next up, we need a page where the user can select what data source they wish to use for their visualisation.

In this issue, you only need to implement the page. You **do not** need to wire it up to a global store for handling selection of data sources.

### Subtasks
- [x] Create data source selection page
- [x] Retrieve data about available data sources from backend using GraphQL
- [x] Create buttons (useless ones) for adding data sources

emilom:
Closes #53
Added organism for retrieving and displaying metaData from backend.
Added molecule for displaying each dataset retrieved from the backend. 

Created two stories for these components and mocked the result with faker.


src/components/atoms/DatasetInfoBox.tsx
```diff
@@ -0,0 +1,26 @@
...
+type DatasetInfoBoxProps = {
+    title: string;
+    description: string;
+    tags: string[];
+};
+
+const DatasetInfoBox: React.FC<DatasetInfoBoxProps> = ({ title, description, tags }) => {
+    return (
+        <Pane width="100%" height="100%">
+            <Heading size={600}>{title}</Heading>
```
blauks: Could we add the heading size as an optional prop? And just have the default size be 600

src/components/organisms/DataExplorer.tsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { Pane } from 'evergreen-ui';
+import DataResultItem from '../molecules/DataResultItem';
+import { useQuery } from '@apollo/client';
+import { METADATA, MetadataEntry } from '../../queries/metadata';
+import Loading from '../atoms/Loading';
+import ErrorMessage from '../atoms/ErrorMessage';
+
+const DataExplorer: React.FC = () => {
+    const { data, loading, error } = useQuery(METADATA);
```
blauks: useQuery should take in the metadata type, because typescript 😃 ```useQuery<T>(METADATA)```

src/mockdata/mocks.ts
```diff
@@ -0,0 +1,88 @@
...
+import { addDays } from 'date-fns';
+import { WEATHER_MET_API } from '../queries/metApi';
+import { METADATA } from '../queries/metadata';
+import faker from 'faker';
+
+const visualisationTypes = ['barchart', 'thresholdchart', 'piechart', 'linechart'];
+
+const randomListOfWords = () => {
+    const wordList = faker.lorem.words(faker.random.number(7)).split(' ');
+    return wordList.length !== 1 ? wordList : [];
```
blauks: Isn't there any other way to do this? Now it isn't possible to fake one tag. Probably not important, it's up to you. Another way is to just check if ```wordList[0] === ""```

src/queries/metadata.ts
```diff
@@ -0,0 +1,53 @@
+import { gql } from '@apollo/client';
+
+//Waiting for backend to merge.
```
blauks:
```suggestion
```

src/stories/DataExplorer.stories.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { Meta } from '@storybook/react/types-6-0';
+import DataExplorer from '../components/organisms/DataExplorer';
+
+export const Primary: React.FC = () => <DataExplorer />;
+
+export default {
+    title: 'Component/DataExplorer',
+    component: DataExplorer,
+} as Meta;
```
blauks:
```suggestion

export default {
    title: 'Component/DataExplorer',
    component: DataExplorer,
} as Meta;

export const Primary: React.FC = () => <DataExplorer />;
```

blauks (commented):
:)

src/mockdata/mocks.ts
emilom: True. Couldn’t find any way to do it with just faker so rushed a quick solution. Didn’t think about the case when only one tag was added. wordList[0] === "" would work.

src/components/atoms/DatasetInfoBox.tsx
emilom: I don't think the headings size will differ in any way on different components. This component will be used on the data explorer page and the page where you can customise and add the visualisation to your dashboard, and the styling will be the same. No problem with adding props with default values tho

src/components/organisms/DataExplorer.tsx
emilom: Smart 👯

src/components/molecules/DataResultItem.tsx
```diff
@@ -0,0 +1,29 @@
...
+    };
+
+    return (
+        <Card width="100%" height="20rem" elevation={1} display="flex" onClick={handleOnClick}>
+            <Pane flex="1" padding="1rem">
+                <DatasetInfoBox title={title} description={description} tags={tags} />
+            </Pane>
+            <Pane flex="2">
+                <Heading>Forslag til visualisering</Heading>
+                {/* The template graph will go here */}
```
blauks, line 24: Would it be possible to place the threshold chart here just as a placeholder? So that we know that the design works. Also as with the DashboardItem, shouldn't the graph be passed in as a child?

FredrikAugust:
Screenshot 2021-03-15 at 10 11 34


Would it be possible to get some more descriptive names in the storybook?
E.g.
- Component header stories could be moved into Dashboard/Dashboard/Primary and "Data Explorer"/... (note that you can have spaces in the story names which looks a bit cleaner (if I'm not mistaken)).
- Rename Dashboard/Item to "Data Block", "Visualisation block" or something along those lines
- It should be obvious what the story is of given it's name

src/components/atoms/DatasetInfoBox.tsx
```diff
@@ -0,0 +1,26 @@
...
+            <Text overflow="auto">{description}</Text>
+            <Pane display="flex" flexWrap="wrap">
+                {tags.map((tag, index) => (
```
FredrikAugust: The index should not be used as a key when mapping. In this case it makes no sense for the tags _not_ to be unique, so use them.

src/components/molecules/DataResultItem.tsx
```diff
@@ -0,0 +1,29 @@
...
+};
+
+const DataResultItem: React.FC<DataResultItemProps> = ({ title, description, tags }) => {
+    const handleOnClick = () => {
+        return 0; //Will route to data customise page later.
+    };
+
+    return (
+        <Card width="100%" height="20rem" elevation={1} display="flex" onClick={handleOnClick}>
+            <Pane flex="1" padding="1rem">
```
FredrikAugust: Perhaps move this padding to the container element?

FredrikAugust: Looks a bit more unified.

src/components/organisms/DataExplorer.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { METADATA, MetadataEntry, Metadata } from '../../queries/metadata';
+import Loading from '../atoms/Loading';
+import ErrorMessage from '../atoms/ErrorMessage';
+
+const DataExplorer: React.FC = () => {
+    const { data, loading, error } = useQuery<Metadata>(METADATA);
+
+    if (loading) {
+        return (
+            <Pane height="100vh">
```
FredrikAugust: Is this necessary? What if there's something above it?

src/components/organisms/DataExplorer.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import { Pane } from 'evergreen-ui';
+import DataResultItem from '../molecules/DataResultItem';
+import { useQuery } from '@apollo/client';
+import { METADATA, MetadataEntry, Metadata } from '../../queries/metadata';
+import Loading from '../atoms/Loading';
+import ErrorMessage from '../atoms/ErrorMessage';
+
+const DataExplorer: React.FC = () => {
+    const { data, loading, error } = useQuery<Metadata>(METADATA);
```
FredrikAugust: Can't you use the wrapper you made earlier here?

You're checking for loading, error and data, which is the same as in DataWrapper.

src/components/molecules/DataResultItem.tsx
emilom: I was thinking about just passing the charttype from metadata as a prop, since we will be faking the data anyways. Isn’t it more natural to handle all this in DataResultItem instead of the DataExplorer?

src/mockdata/mocks.ts
```diff
@@ -0,0 +1,88 @@
...
+
+const visualisationTypes = ['barchart', 'thresholdchart', 'piechart', 'linechart'];
+
+const randomListOfWords = () => {
+    const wordList = faker.lorem.words(faker.random.number(7)).split(' ');
+    return wordList[0] ? wordList : [];
+};
+
+const makeMetadata = () => {
+    const allMetadata = [];
```
FredrikAugust: Shouldn't this be typed?

FredrikAugust: E.g. as a `Metadata` type or similar.

src/queries/metadata.ts
```diff
@@ -0,0 +1,52 @@
...
+            axes: {
+                x: {
+                    name: string;
+                };
+                y: {
+                    name: string;
+                };
+            };
+        },
+    ];
```
FredrikAugust: This is a common error which can give you some very funky behaviour later on.

What you're saying here is that `visusalisations` is an array of one (not zero, not more) objects of the type you nested in the array.

What you're looking for is either `Array<{ ... }>` or `{ ... }[]`. This is one reason for why some opt for the former, as it looks less like the mishap here.

src/queries/metadata.ts
```diff
@@ -0,0 +1,52 @@
...
+};
+
+export type Metadata = {
```
FredrikAugust: This seems a bit strange. Metadata to me sounds more like what `allMetadata` returns an array of.

Could you perhaps give it a better name? What about just `AllMetadataResult` or something along those lines?

FredrikAugust (changes requested):

src/components/atoms/DatasetInfoBox.tsx
emilom: good idea 💡

src/components/molecules/DataResultItem.tsx
FredrikAugust: I'd agree with Emil here, since you pass all required information to create a mocked chart from the metadata query, you can generate it here.

blauks: It depends on where we think it is natural to "create" the chart. I suppose it makes more sense to "create" it in the Item, I just didn't see anywhere where the DataExplorer passed this information downwards.

FredrikAugust: It's not passed in here, but it can be added when needed.

https://github.com/IT2901-SINTEF01/frontend/pull/62/files/fb0eb14a0e0d4e5e90c2c8c4b220290b37f36f46#diff-64daaaac0ccbbb9d38bd54d6d882e67c8784385bf68c39acc98085efb830b963R30-R34

src/components/organisms/DataExplorer.tsx
emilom: I guess you could, but we won’t t be formatting any data like we do in dataWrapper

src/components/molecules/DataResultItem.tsx
FredrikAugust: The information is already there, it's just not included in the proplist.

To make this easier @emilom, you could also pass the whole `el` as a prop to `DataResultItem` instead of destructuring it and extracting only a subset of fields.

FredrikAugust: I.e.

```tsx
<DataResultItem metadata={el} />
```

src/components/organisms/DataExplorer.tsx
FredrikAugust: Then you could simply have the mapping function be `e => e`.

src/queries/metadata.ts
emilom: Good catch 🤽🏽‍♂️

src/components/organisms/DataExplorer.tsx
```diff
@@ -0,0 +1,40 @@
...
+        return (
+            <Pane height="100vh">
+                <ErrorMessage message="Det oppsto en feil under henting av datasett" moreInfo={error.message} />
+            </Pane>
+        );
+    }
+
+    return (
+        <Pane>
+            {data &&
```
blauks: If data is undefined will the page just be blank? Maybe there could be a message (ish like the Error message) saying something like "Ingen data funnet🤒" or something.

src/components/organisms/DataExplorer.tsx
FredrikAugust: You will probably have to type the parameters, however, as using generic parameters on the component seems to break your current build CI action.

E.g.

```tsx
(e: Type): Type => e
```

I would argue that it would be better to include the generic parameters on the component, but this requires @blauks to look at the CI action and resolve the issues there.

src/components/organisms/DataExplorer.tsx
FredrikAugust: FWIW, this is handled if you opt for using the `DataWrapper`, ref.

```tsx
    if (!data) {
        return <ErrorMessage message="Ingen data mottatt fra serveren." />;
    }
```

src/components/organisms/DataExplorer.tsx
blauks: It only breaks the Docker CI, which I do not understand 😀

FredrikAugust: You're not expected to understand it now, but it probably has a logical explanation.

The docker configuration is deterministic, so it should in all theory be reproducible locally -- and thus easier to debug and fix.

FredrikAugust: The CI part shouldn't affect the outcome -- as is the point of docker.

src/components/organisms/DataExplorer.tsx
emilom: I thought it was to get the loading icon centred because i was using storybook.
It should probably be changed to just 100%.

src/components/organisms/DataExplorer.tsx
emilom: DataExplorer will be expanded with search and filter later. Isn’t it better to let DataExplorer handle all this instead of dataWrapper? The styling will also be different for errors and loading so DataWrapper must be changed if we want to use it in DataExplorer.

src/components/organisms/DataExplorer.tsx
emilom: I dont think this should be an error. Is more feedback to the user that nothing was found for the given search input or filters once we implement these features.

src/components/organisms/DataExplorer.tsx
blauks: I agree, I think it is best to keep them seperated.

src/components/molecules/DataResultItem.tsx
emilom: I didn’t add ThresholdChart as a placeholder now.  This can be added in a later issue when we implement the visualisation proposals.

src/index.css
```diff
@@ -5,4 +5,9 @@ html {
 
 body {
     margin: 0;
+    height: 100%;
 }
+
+#root {
+    height: 100%;
+}
```
blauks: ?

emilom: It is needed for positioning and styling.

emilom: the root div had a fixed height, so it wasn’t possible to center vertically

src/stories/DataResultItem.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+export const Primary = Template.bind({});
+Primary.args = {
+    title: 'This is an title',
+    description:
+        'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.',
+    tags: ['Energi', 'Miljø', 'Helse', 'Pils', 'Raketter'],
+    visualisationType: 'thresholdchart',
+};
+
+Primary.storyName = 'Data explorer result item';
```
blauks: Couldn't this just be called ```Item```, to keep the naming convention the same as the Dashboard. So you could edit line 7 to be ```title: 'DataExplorer/DataExplorer',```, and this to ```Item```.

src/stories/DataResultItem.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+} as Meta;
+
+const Template: Story<DataResultItemProps> = (args) => <DataResultItem {...args} />;
+
+export const Primary = Template.bind({});
+Primary.args = {
+    title: 'This is an title',
+    description:
+        'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.',
+    tags: ['Energi', 'Miljø', 'Helse', 'Pils', 'Raketter'],
```
blauks, line 18: The tags should be able to edited/controlled without typing in a string, I would recommend looking into the multiselect control:D https://storybook.js.org/docs/react/essentials/controls#annotation

blauks (commented):
I also noticed that the text and title sizes are larger than on the items on the Dashboard, is this on purpose?

src/stories/DataResultItem.stories.tsx
blauks: Also one should be able to choose the visualisation type of an item from a predefined set of types, this could be "controlled" with a select control

src/index.css
blauks:
```suggestion
}

```

blauks (dismissed):
LGTM TO ME

src/components/organisms/DataExplorer.tsx
```diff
@@ -0,0 +1,55 @@
...
+        return (
+            <Pane height="100%" display="flex" justifyContent="center" alignItems="center">
+                <InlineAlert intent="warning">Vi fant dessverre ingen datasett knyttet til ditt søk.</InlineAlert>
+            </Pane>
+        );
+    }
+
+    return (
+        <Pane>
+            {data.allMetadata.map((el: MetadataEntry) => (
```
FredrikAugust:
```suggestion
            {data.allMetadata.map((el) => (
```

src/index.css
```diff
@@ -5,4 +5,9 @@ html {
 
 body {
     margin: 0;
+    height: 100%;
+}
+
+#root {
+    height: 100%;
```
FredrikAugust:
```suggestion
    min-height: 100%;
```

src/index.css
```diff
@@ -5,4 +5,9 @@ html {
 
 body {
     margin: 0;
+    height: 100%;
```
FredrikAugust:
```suggestion
    min-height: 100%;
```

src/queries/metadata.ts
```diff
@@ -0,0 +1,51 @@
...
+        type: string;
+        axes: {
+            x: {
+                name: string;
+            };
+            y: {
+                name: string;
+            };
+        };
+    }>;
```
FredrikAugust:
```suggestion
    visualisations: {
        type: string;
        axes: {
            x: {
                name: string;
            };
            y: {
                name: string;
            };
        };
    }[];
```

FredrikAugust (changes requested):

FredrikAugust (approved):
LOOGIETOOMIE!

PR #68, opened 2021-03-18
Test apollo state manager

### LINKED ISSUE #55: Global store for keeping track of selected visualisations
### Description
The frontend needs to keep track of which visualisations are selected, and a global store is a suitable place for that. I believe @emilom found the Apollo state management to be sufficient, so that should be used.

### Subtasks
- [x] Create store and model
- [ ] **Test** (Unit test) selecting a data source and make sure it's reflected in the store. Do not need UI test

blauks:
Closes #55 

This PR makes it possible to test our react components individually, and I also added two tests to check that the Apollo state manager behaves like it supposed to.


cypress/component/unit_test_apollo_state_updates.spec.tsx
```diff
@@ -0,0 +1,21 @@
...
+describe('Apollo state management', () => {
+    beforeEach(() => {
+        dashboardItemsVar([]);
+    });
+    it('dashboard starts out empty', () => {
+        mount(<Dashboard />);
+        expect(dashboardItemsVar()).to.deep.equal([]);
+    });
+    it('adds new item to state', () => {
+        // This has to change when we add more functionality to the "Add button" button
```
emilom: Add button button?

emilom (commented):
Everything else LGTM to me

cypress/component/unit_test_apollo_state_updates.spec.tsx
blauks: Sorry bud, I will run grammarly next time. Good spot👯‍♀️

cypress/component/unit_test_apollo_state_updates.spec.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { mount } from '@cypress/react';
+import Dashboard from '../../src/components/organisms/Dashboard';
+import AddComponentButton from '../../src/components/atoms/AddComponentButton';
+import { dashboardItemsVar } from '../../src/cache/index';
+
+describe('Apollo state management', () => {
+    beforeEach(() => {
+        dashboardItemsVar([]);
```
FredrikAugust: What does invoking this do? Wouldn't it be better to create this variable each time and pass it into the `it` blocks? That way you can ensure you don't encounter any race conditions if e.g. one test runs before the other and mutates the variable?

cypress/component/unit_test_apollo_state_updates.spec.tsx
emilom: Talk to @thomran 🤓

cypress/component/unit_test_apollo_state_updates.spec.tsx
FredrikAugust: I.e. use `makeVar<DashboardItemInfo[]>([]);` in the `beforeEach` to create a fresh state each test.

FredrikAugust (changes requested):

cypress/component/unit_test_apollo_state_updates.spec.tsx
blauks: That is very true, very good point.

blauks: Hey, me again. Seems like this is easier said than done. We use this variable multiple places in the application (it's just how apollo state manager works), so we cannot make a new variable just for these tests, and we cannot really use makeVar to initialise a fresh state since the variable is imported. So, with my restricted knowledge, I would say that just setting the variable to [] before each test is adequate enough. If this causes any problems down the line, I can owe you a small beverage.

FredrikAugust: Alright. Looking forward to the small beverage (or not).

FredrikAugust:
> Everything else LGTM to me

How do you pronounce this?

FredrikAugust (approved):

PR #78, opened 2021-03-21
Deep comparison

### LINKED ISSUE #70: Use deep object comparison in cypress test
https://github.com/IT2901-SINTEF01/frontend/blob/cc486cd267b270c266339bc308f4c8d74cebcc62/cypress/component/unit_test_apollo_state_updates.spec.tsx#L28

Right now it appears to be using shallow object comparison which means that it will always pass.

blauks:
Closes #70 

I updated the test a little bit, and made use of a deep compare instead of a shallow(?) one.

### DIFF
--- a/cypress/component/unit_test_apollo_state_updates.spec.tsx
+++ b/cypress/component/unit_test_apollo_state_updates.spec.tsx
@@ -1,9 +1,13 @@
 import React from 'react';
 import { mount } from '@cypress/react';
-import Dashboard from '../../src/components/organisms/Dashboard';
-import AddComponentButton from '../../src/components/atoms/AddComponentButton';
-import { dashboardItemsVar } from '../../src/cache/index';
 import { BrowserRouter } from 'react-router-dom';
+import { dashboardItemsVar } from '../../src/cache/index';
+import { MockedProvider } from '@apollo/client/testing';
+import mocks from '../../src/mockdata/mocks';
+import { cache } from '../../src/cache/index';
+import { WEATHER_MET_API } from '../../src/queries/metApi';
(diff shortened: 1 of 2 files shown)


emilom (approved):
LGTM

cypress/integration/website.spec.tsx
```diff
@@ -3,3 +3,5 @@ describe('Dashboard', () => {
         cy.visit('/');
     });
 });
+
+export {};
```
FredrikAugust, line 7: 0d8e3ead368182625fc060808ed0ea1cb0687cc6

This can also be achieved without exporting by updating the `tsconfig` to support `isolatedModules`.

Source: https://www.typescriptlang.org/tsconfig#isolatedModules

cypress/component/unit_test_apollo_state_updates.spec.tsx
```diff
@@ -17,14 +21,21 @@ describe('Apollo state management', () => {
...
-        cy.contains('Legg til nytt komponent').click();
-        expect(dashboardItemsVar()).to.not.equal([]);
+        const item = {
+            title: 'This is a custom title',
+            id: 'This is a custom ID',
+            size: DashboardItemSize.SMALL,
+            query: WEATHER_MET_API,
+        };
+        dashboardItemsVar([item]);
+        expect(dashboardItemsVar()).to.deep.include(item);
```
FredrikAugust, line 39: This doesn't test our business logic, only if the library we're using works.

blauks: 👨‍💼

PR #81, opened 2021-03-22
Use query matching selected data source on dashboard

### LINKED ISSUE #71: Correct GQL Query should be used on dashboard
### Description

Right now, the query used on the dashboard is not correct, and this needs to change.

We are gathering an army and preparing to tackle this enormous hindrance.

Sharpen your tools, we're going into war.

### Subtasks
- [x] Create function for determining which type and query should be used based on selections on data chooser pages

FredrikAugust:
Closes #71 

Changelog
-----------
- The query and mapping functions are now decided based on a function on the data source name (from backend)

Notes
------
- The `Chart` type is currently not based on this information


emilom (dismissed):
Lookin good

src/mockdata/mocks.ts
```diff
@@ -18,7 +19,7 @@ export const makeMetadata = () => {
     for (let i = 0; i < 10; i++) {
         allMetadata.push({
             id: faker.random.uuid(),
-            name: faker.name.title(),
+            name: DataSourceName.MET_API_FORECAST,
```
blauks, line 22: Was this change necessary?

FredrikAugust: Yes. 

I am, however, in the process of creating issues which will remove the need for this.

The reason for why they are there now is because the title is used to determine which query should be used to retrieve data from the backend. The type of the `name` field is no longer only a string, but actually an enum with the different names available in the DB.

This is error-prone, and thus I'm proposing a migration to using crafted IDs like `SSB_POPULATION` or `MET_API_FORECAST` for the metadata entries.

Also, refer to #83 -- though this will likely be made obsolete by the proposed changes.

blauks: Sounds good and probably also smart

PR #87, opened 2021-03-24
Create line chart component

### LINKED ISSUE #79: Create new visualisation component for populationsInNorway data set
### Description
A threshold chart will not be the best way to visualise population data in Norway. We should create a new component for line charts which can be added to our library of available visualisation methods.

### Subtasks
- [ ] Create a line chart in visx

emilom:
Close #79 

Added LineChart component and created a new type in types/visualisation which can be used by all visualisations.


blauks (approved):
Loogietoomie

PR #88, opened 2021-03-24
New E2E tests :D

### LINKED ISSUE #57: Test adding data visualisations with Cypress
### Description
Wow! A complete e2e test!

You should test that adding a data visualisation and navigating between the relevant pages works as expected.

### Subtasks
- [x] Test navigation with 404s as well
- [x] Test adding visualisation
- [x] Test deleting visualisation

blauks:
Closes #57 

I have mocked the graphql requests sent in cypress, this might not be the best way to do it but it works. I also added a test to check that adding new items to the dashboard works, and one for checking that removing an item works. Additionally, I added a test to check that the application sends you to a 404 page when one tries to go to a random sub-page.

### DIFF
--- a/src/components/organisms/Dashboard.tsx
+++ b/src/components/organisms/Dashboard.tsx
@@ -52,7 +52,7 @@ const Dashboard: React.FC = () => {
                             width="100%"
                             height="100%"
                             gridColumn={'span ' + item.size}
-                            className={'dashboardItem'}
+                            className="dashboardItem"
                         >
                             <DashboardItem
                                 title={item.name}
--- a/src/components/organisms/Dashboard.tsx
+++ b/src/components/organisms/Dashboard.tsx
@@ -47,7 +47,13 @@ const Dashboard: React.FC = () => {
             >
                 {dashboardItems.map((item) => {
                     return (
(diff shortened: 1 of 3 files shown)


src/components/organisms/Dashboard.tsx
```diff
@@ -47,7 +47,13 @@ const Dashboard: React.FC = () => {
             >
                 {dashboardItems.map((item) => {
                     return (
-                        <Pane key={item.id} width="100%" height="100%" gridColumn={'span ' + item.size}>
+                        <Pane
+                            key={item.id}
+                            width="100%"
+                            height="100%"
+                            gridColumn={'span ' + item.size}
+                            className={'dashboardItem'}
```
emilom:
```suggestion
                            className="dashboardItem"
```

emilom (dismissed):
Looks good, maybe change classname to just string?

FredrikAugust (dismissed):

src/components/organisms/Dashboard.tsx
blauks: Ok Mr. React

FredrikAugust (approved):

PR #94, opened 2021-03-26
Use random data source name

### LINKED ISSUE #83: Mock should return random valid name for data source
### Description
Mocks only return one title for the visualisations, while it should return a sample (read: _one_) of all available data sources for _randomness_.

### Steps to reproduce
1. Look at the code and run it and then look at the results

### Expected behaviour
Very random (:sparkles: _quirky_ :sparkles:) selection.

Kpaubert:
Closes #83 
Faker is now used to get a random API-name from the mocking service instead of just using a single one.


blauks (approved):
Looks LGTM to me

FredrikAugust (approved):

PR #95, opened 2021-03-26
Display available visualisations on the visualisation selection page

### LINKED ISSUE #84: Display available visualisations on the visualisation selection page
### Description
This is actually a bit more work than illustrated by the title.

The backend metadata field returns a set of visualisations which have a `type` field, which we can map to different component types (e.g. `ThresholdChart`, `BarChart`, &c).

We need to use this information, alongside the information regarding what data source we are working with to retrieve the correct `mappingFunction`, map the data to the correct format, and inject it into the visualisation. This requires some changes to the way the mapping function is retrieved today.

With this issue (and corresponding PR) you should be able to view the different visualisations available to the data source.

#### Some considerations
- The `visualisation` object from the backend includes some metadata which should be passed to the visualisation (e.g. `name`, `axesName` &c)

### Subtasks
- [ ] Map `visualisation['type']` to chart type
- [ ] Map `visualisation['type']` to mapping function (with data source added to the equation)
- [ ] Generate "faked" visualisations on the selection page

### Additional information
This issue does not include the actual mapping of real data to the different visualisations. That will be handled in another issue. This does however lay the foundation for that.

FredrikAugust:
Closes #84 

Changelog
-----------

- Generate visualisations based on metadata
- Show visualisations on visualisation editor page
- Create faker for `TimeEntry[]`


src/components/atoms/VisusalisationsSelector.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { SelectField } from 'evergreen-ui';
+
+type Props = {
+    label: string;
+    options: Record<string, string>;
+    onChange: React.ChangeEventHandler<HTMLSelectElement>;
+};
+
+const VisualisationsSelector: React.FC<Props> = (props) => (
+    <SelectField label={props.label} defaultValue={Object.values(props.options)[0]} onChange={props.onChange}>
```
Kpaubert: Perhaps make a variable for `Object.values(props.options)[0]` making it easier to see what this value means.

src/components/atoms/VisusalisationsSelector.tsx
```diff
@@ -0,0 +1,20 @@
...
+        {Object.entries(props.options).map(([label, value]) => (
```
Kpaubert: ^

src/utils/visualisationLabels.ts
```diff
@@ -0,0 +1,10 @@
+import { VisualisationType } from '../types/Metadata';
+
+export function friendlyNameForVisualisationType(type: VisualisationType): string {
+    return {
+        [VisualisationType.LINE]: 'Linjegraf',
+        [VisualisationType.BAR]: 'Bargraf',
```
Kpaubert: "Søylediagram" rather than "Bargraf".

src/mockdata/mocks.ts
```diff
@@ -4,6 +4,7 @@ import { WEATHER_MET_API } from '../queries/metApi';
 import { METADATA, MetadataEntry } from '../queries/metadata';
 import faker from 'faker';
 import { DataSourceName } from '../utils/dataSourceMappings';
+import { VisualisationType } from '../types/Metadata';
 
 faker.seed(42069);
```
Kpaubert: I don't really appreciate these kinds of connotations.

Kpaubert (commented):
Looks good, just need some clarifications.

src/components/atoms/VisusalisationsSelector.tsx
FredrikAugust: Post-discussed deemed not necessary.

Only used once.

Takes first value in map.

src/components/atoms/VisusalisationsSelector.tsx
FredrikAugust: Post-discussed deemed not necessary.

Only used once. Iterates over Map.

src/mockdata/mocks.ts
FredrikAugust: Ok/
Same

FredrikAugust:
Seems to be an issue with GCR. Docker images build fine normally.

src/components/atoms/VisusalisationsSelector.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { SelectField } from 'evergreen-ui';
+
+type Props = {
+    label: string;
+    options: Record<string, string>;
```
blauks: Question:
Is this the same as typing ```options: {x: string, y: string}```?

src/mockdata/mockTimeEntry.ts
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * @param size Size of the returned array.
+ * @param min Minimum value of time entry value.
+ * @param max Maximum value of time entry value.
+ */
+export default function (size: number, [min, max]: [number, number] = [0, 100]): TimeEntry[] {
+    const values = new Array(size).fill(null).map(() => Math.random() * (max - min) + min);
+    const timeseries = Object.keys(new Array(size).fill(null)).map((i) => addDays(new Date(), Number(i)));
+    return values.map((e, i) => ({ value: e, time: timeseries[i].toISOString() }));
+}
```
blauks: Instead of this, couldn't you use the the makeTimeseries function that already exists in mocks.ts? And just map that to TimeEntry with the mapping function?

src/utils/visualisationLabels.ts
```diff
@@ -0,0 +1,10 @@
+import { VisualisationType } from '../types/Metadata';
+
+export function friendlyNameForVisualisationType(type: VisualisationType): string {
+    return {
+        [VisualisationType.LINE]: 'Linjediagram',
+        [VisualisationType.BAR]: 'Søylediagram',
+        [VisualisationType.PIE]: 'Kakediagram',
```
blauks: ```Sektordiagram``` would sound more fancy:-)

blauks (dismissed):
Looks mostly good to me! Just sceptical to adding the mockTimeEntry file. You could just export weather data from the mocks.ts file, and format it with the mapping function. But if you feel like this file will make our lives more easy, we can just keep it.

blauks:
> Seems to be an issue with GCR. Docker images build fine normally.

Is this something we can fix?

FredrikAugust:
> > Seems to be an issue with GCR. Docker images build fine normally.
> 
> 
> 
> Is this something we can fix?

Nope. There was an issue with the docker registry. Should work now.

src/components/atoms/VisusalisationsSelector.tsx
FredrikAugust: No, not quite. 

`Record<string, string>` is an object where both the keys and the values are strings. Your example would conform to this type, but it would be a subset. 

An alternative typing would be (IIRC, typing from mobile):

`{ [key: string]: string }`

FredrikAugust: Record is very useful when you know the type of the object keys and values, but not the exact key names.

blauks: Ah ok, looks smart atleast!

src/components/atoms/VisusalisationsSelector.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
```
blauks: ```VisusalisationsSelector.tsx```
Not to be Mr. Grammarly or anything, but I think there is something fishy🐟 about the filename. I suggest changing it to VisualisationSelector.

src/mockdata/mockTimeEntry.ts
FredrikAugust: [This](https://github.com/IT2901-SINTEF01/frontend/blob/c9790fdf5f095baedb29415c57f42b8ce54fa363/src/mockdata/mocks.ts#L46) function creates a timeseries object for use with forecast data. The name `makeTimeseries` might be slightly misleading as it's meant for only one use case (or so it seems).

src/components/atoms/VisusalisationSelector.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
```
blauks: Can we change the filename to ```VisualisationSelector``` aswell?

src/mockdata/mockTimeEntry.ts
```diff
@@ -0,0 +1,15 @@
+import { TimeEntry } from '../types/visualisation';
```
blauks: Will the mocked data be random everytime we run this? If it is, we will end up with the same bug as we had a couple weeks ago where the UI-tests will think that the design has been updated every time you re-render the graphs using this as a data source, which would not be fun. Is there a way to make this data predictable?

blauks (commented):
Looks good! We just need to find a way to make the mockTimeEntry file to be predictable.

src/mockdata/mockTimeEntry.ts
FredrikAugust: No.

FredrikAugust: I will implement a fix now.

FredrikAugust: Resolved.

blauks (dismissed):
_L o o g y_  __Too mey!__ (british)

But R.I.P. to my funny faker seed :(

blauks (dismissed):
Still looks LGTM to me!

Kpaubert (approved):
LGTM! 🔥 🧯

PR #99, opened 2021-04-03
Add support for numbers on x-axis of charts

### LINKED ISSUE #92: Generalise ThresholdChart and LineChart to work with non-date X-axis values
### Description
As the title reads.

Both should support non-date values as well.

Take a look at the metadata object to see the available values.

Both should also still support using time entries.

### Subtasks
- [ ] Update ThresholdChart
- [ ] Update LineChart

emilom:
LineChart and ThresholdChart now works with numbers and dates on x axis instead of just dates.
Closes #92


src/components/atoms/ThresholdChart.tsx
```diff
@@ -29,59 +29,64 @@ const ThresholdChart: React.FC<ThresholdChartProps> = ({
...
 
-    const valueScale = scaleLinear({
+    const yScale = scaleLinear({
         domain: [minValue, maxValue],
         range: [yMax, 0],
         nice: true, //Gives the graph more space, rounds the y axis to closest integer.
     });
 
-    const firstDate = date(data[0]);
+    const firstX = xValue(data[0]);
```
blauks, line 64: Do we need to make this a variable?

blauks (approved):
Looks good:D Just wondering if you need to make firstX a variable, since you only use it once.

src/components/atoms/ThresholdChart.tsx
emilom: Not really. I guess it makes the code more readable in some way?

PR #103, opened 2021-04-06
Improve accessibility

### LINKED ISSUE #89: Improve accessibility
A review of the application accessibility can be found [here](https://www.notion.so/Accessibility-review-f199dd0e4aec4a0f9a4f227b97cadf8c). The review provides an overview of tasks to be done and explains the tasks below in greater detail.

- [x] Make all custom controls into native HTML elements. If this is not possible, they should have associated labels and ARIA roles.
- [x] Add labels and/or text alternatives where necessary.
- [x] Give focus to pop-up alerts when they appear.

emilom:
Attempt at improving accessibility. 
I didn’t understand all the points mentioned in #89 like "The dropdown div(s) and checkbox div should be changed to be native HTML element(s)" since most of the evergreen elements are just styled native HTML elements. 

The dashboard and data explorer are updated with ARIA roles. I removed the switch from data visualisation editor as it made accessibility difficult and serve no real purpose since an empty textarea gives the same result. 

Please give some feedback @thomran if there’s something i'm not understanding about accessibility 😄. 

and i removed the footer... 

Closes #89


src/components/atoms/VisualisationParameterSelector.tsx
```diff
@@ -44,20 +31,13 @@ const VisualisationParameterSelector: React.FC<VisualisationParameterSelectorPro
...
-                    </Text>
-                )}
-            </Pane>
+            <Label htmlFor="textarea" marginBottom={12}>
+                Legg til text
+            </Label>
             <Pane flex="1">
                 <Textarea
-                    disabled={!withText}
+                    id="textarea"
```
FredrikAugust: Perhaps attempt to use more descriptive IDs. This could quickly cause collisions:)

src/components/molecules/DataResultItem.tsx
```diff
@@ -10,12 +10,17 @@ export type DataResultItemProps = {
...
-        <Card width="100%" height="20rem" elevation={1} display="flex" onClick={handleOnClick} padding="1rem">
+        <Card
+            width="100%"
+            height="20rem"
+            elevation={1}
+            display="flex"
+            padding="1rem"
+            role="button"
+            aria-pressed="false"
+            aria-label={`Gå til visualiserings velger for ${title}.`}
```
FredrikAugust:
```suggestion
            aria-label={`Gå til visualiseringsvelger for ${title}.`}
```

særskrivingsfeil

FredrikAugust (changes requested):

src/components/atoms/VisualisationParameterSelector.tsx
emilom: true

FredrikAugust: !false

emilom: true + true === 2

thomran (commented):
Looks good! As the switch to enable text within a component is now removed (and I agree, it was redundant), perhaps it would be a good idea to change the "title" in order to make it clear that the text input field is optional. Maybe it could be changed from "Legg til tekst" to "Legg til tekst _(valgfritt)_" or something similar?

FredrikAugust (approved):

PR #108, opened 2021-04-07
Persist choices with localstorage

### LINKED ISSUE #107: Persist choices
### Description
When the user refreshes the page, the selected visualisations should stay.

### Subtasks
- [x] Store changes in localstorage through a typed serialised JSON object
- [x] Retrieve localstorage cache if present on page load

emilom:
The apollo reactive var will now be initialised with data from localstorage if it exists.  The add/remove item button updates localstorage with updated values. 

Close #107

### DIFF
--- a/src/cache/index.ts
+++ b/src/cache/index.ts
@@ -15,4 +15,9 @@ export const cache = new InMemoryCache({
     },
 });
 
-export const dashboardItemsVar = makeVar<DashboardItemInfo[]>([]);
+const initValue = [];
+const storage = window.localStorage.getItem('dashboard');
+if (storage) {
+    initValue.push(...JSON.parse(storage));
+}
+export const dashboardItemsVar = makeVar<DashboardItemInfo[]>(initValue);
--- a/src/cache/index.ts
+++ b/src/cache/index.ts
@@ -15,6 +15,10 @@ export const cache = new InMemoryCache({
     },
 });
 
+export const updateLocalstorage = (saveData: DashboardItemInfo[]) => {
+    window.localStorage.setItem('dashboard', JSON.stringify(saveData));
+};
+
 const storage = window.localStorage.getItem('dashboard');
 const initValue: DashboardItemInfo[] = storage ? JSON.parse(storage) : [];
 export const dashboardItemsVar = makeVar<DashboardItemInfo[]>(initValue);
--- a/src/cache/index.ts
+++ b/src/cache/index.ts
@@ -15,9 +15,6 @@ export const cache = new InMemoryCache({
     },
 });
 
-const initValue = [];
 const storage = window.localStorage.getItem('dashboard');
-if (storage) {
-    initValue.push(...JSON.parse(storage));
-}
+const initValue: DashboardItemInfo[] = storage ? JSON.parse(storage) : [];
 export const dashboardItemsVar = makeVar<DashboardItemInfo[]>(initValue);
--- a/src/cache/index.ts
+++ b/src/cache/index.ts
@@ -15,7 +15,7 @@ export const cache = new InMemoryCache({
     },
 });
 
-export const updateLocalstorage = (saveData: DashboardItemInfo[]) => {
+export const updateLocalstorage = (saveData: DashboardItemInfo[]): void => {
     window.localStorage.setItem('dashboard', JSON.stringify(saveData));
 };
 
--- a/src/components/atoms/VisualisationParameterSelector.tsx
+++ b/src/components/atoms/VisualisationParameterSelector.tsx
@@ -3,11 +3,16 @@ import { Pane, SelectField, Textarea, Label } from 'evergreen-ui';
 import { DashboardItemSize } from '../../types/dashboard';
 
 type VisualisationParameterSelectorProps = {
+    size: DashboardItemSize;
     setSize: (size: DashboardItemSize) => void;
     setParagraph: (text: string | undefined) => void;
 };
 
-const VisualisationParameterSelector: React.FC<VisualisationParameterSelectorProps> = ({ setSize, setParagraph }) => {
+const VisualisationParameterSelector: React.FC<VisualisationParameterSelectorProps> = ({
+    size,
+    setSize,
+    setParagraph,
+}) => {
     const handleSizeChange: React.ChangeEventHandler<HTMLSelectElement> = (e) => {
         switch (e.currentTarget.value as 'small' | 'medium' | 'large') {
             case 'small':
@@ -19,6 +24,17 @@ const VisualisationParameterSelector: React.FC<VisualisationParameterSelectorPro
         }
     };
 
+    const defaultValue = () => {
+        switch (size) {
+            case DashboardItemSize.LARGE:
+                return 'large';
+            case DashboardItemSize.MEDIUM:
+                return 'medium';
+            case DashboardItemSize.SMALL:
+                return 'small';
+        }
+    };
+
     const handleTextChange = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
         setParagraph(e.currentTarget.value);
     };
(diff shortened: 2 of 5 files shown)


src/components/molecules/AddToDashboard.tsx
```diff
@@ -14,11 +14,15 @@ const AddToDashboard: React.FC<AddToDashboardProps> = (props) => {
...
     };
 
     const removeDashboardItem = () => {
-        dashboardItemsVar(dashboardItemsVar().filter((el) => el.id !== props.dashboardItemInfo.id));
+        const dashboardContent = dashboardItemsVar().filter((el) => el.id !== props.dashboardItemInfo.id);
+        dashboardItemsVar(dashboardContent);
+        window.localStorage.setItem('dashboard', JSON.stringify(dashboardContent));
     };
 
     return (
```
FredrikAugust: Can you use a use-effect which triggers on changes to the variable instead of adding this everywhere you modify the variable?

src/cache/index.ts
```diff
@@ -15,4 +15,9 @@ export const cache = new InMemoryCache({
     },
 });
 
-export const dashboardItemsVar = makeVar<DashboardItemInfo[]>([]);
+const initValue = [];
+const storage = window.localStorage.getItem('dashboard');
+if (storage) {
+    initValue.push(...JSON.parse(storage));
+}
```
FredrikAugust: Use types!

src/cache/index.ts
```diff
@@ -15,4 +15,9 @@ export const cache = new InMemoryCache({
     },
 });
 
-export const dashboardItemsVar = makeVar<DashboardItemInfo[]>([]);
+const initValue = [];
+const storage = window.localStorage.getItem('dashboard');
+if (storage) {
+    initValue.push(...JSON.parse(storage));
```
FredrikAugust: Could you `let initValue: Type = []` and simply assign instead? Would be a bit easier to understand.

FredrikAugust (changes requested):

src/components/molecules/AddToDashboard.tsx
emilom: I changed it a bit now

emilom:
I made it som the parameters are saved on the visualisation editor page as well. Size and the paragraph are saved but the visualisation is not stored now so this must be added later.

src/components/molecules/AddToDashboard.tsx
```diff
@@ -10,19 +10,19 @@ type AddToDashboardProps = {
...
 const AddToDashboard: React.FC<AddToDashboardProps> = (props) => {
     const dashboardItems = useReactiveVar(dashboardItemsVar);
-
     const added = useMemo(() => dashboardItems.some((el) => el.id === props.dashboardItemInfo.id), [dashboardItems]);
 
+    const updateStorage = (content: DashboardItemInfo[]) => {
+        dashboardItemsVar(content);
+        window.localStorage.setItem('dashboard', JSON.stringify(content));
+    };
+
```
FredrikAugust: I would prefer if the localstorage modification was made somewhere else in the application with the use of a useEffect hook.

This way you could delete the visualisation from the dashboard and the localstorage entry would update "automatically", instead of either
a) it being updated the next time someone clicks this button
b) updated because someone added this code to another place (decreasing DRY-ness)

FredrikAugust (changes requested):

src/components/atoms/VisualisationParameterSelector.tsx
```diff
@@ -19,14 +26,25 @@ const VisualisationParameterSelector: React.FC<VisualisationParameterSelectorPro
...
     };
 
     return (
         <Pane height="100%" display="flex" flexDirection="column">
             <Pane height="32px" marginBottom="1rem" />
-            <SelectField label="Velg størrelse" onChange={handleSizeChange} defaultValue="large">
+            <SelectField label="Velg størrelse" onChange={handleSizeChange} defaultValue={defaultValue()}>
                 <option value="small">Liten</option>
                 <option value="medium">Medium</option>
                 <option value="large">Stor</option>
```
blauks: Just a question regarding the options elements. Is it possible to set the option value to the DashboardItemSize enum? This feels a little untidy.

src/components/molecules/VisualisationPreview.tsx
```diff
@@ -21,6 +23,16 @@ type VisualisationPreviewProps = {
...
+    const dashboardItems = useReactiveVar(dashboardItemsVar);
+
+    useEffect(() => {
+        if (!dashboardItems) return;
+        const item = dashboardItems.find((el) => el.id === metadata.id);
+        if (item?.size) {
+            setSize(item.size);
+        }
+        setParagraph(item?.paragraph);
+    }, []);
```
blauks: Shouldn't dashboardItems be inside the []?

blauks (commented):
Looks good to me, just some questions :-)

src/components/atoms/VisualisationParameterSelector.tsx
emilom: Probably. I agree

src/components/molecules/VisualisationPreview.tsx
emilom: yup, thanks 🙏

src/components/molecules/AddToDashboard.tsx
emilom: Not sure if i'm understanding correctly. 
Should i create a function in utils or cache which updates the localstorage? 

Something like this?
```
useEffect(() => {
        updateLocalStorage(dashboardItems) //Some imported function
    }, [dashboardItems]);

    const addDashboardItem = () => {
        dashboardItemsVar([...dashboardItemsVar(), props.dashboardItemInfo]);
    };

    const removeDashboardItem = () => {
        dashboardItemsVar(dashboardItemsVar().filter((el) => el.id !== props.dashboardItemInfo.id));
    };

Comment (FredrikAugust) [reply src/components/molecules/AddToDashboard.tsx]: Yes! Just don't put the useEffect in the molecule. I believe it would make sense to have this at the App.tsx (appx.).

Comment (FredrikAugust) [reply src/components/molecules/AddToDashboard.tsx]: And when using code blocks, you can get syntax highlighting by typing the file extension after the three backticks;)

File: src/components/atoms/VisualisationParameterSelector.tsx (hunk 5 above)
Comment (FredrikAugust) [reply src/components/atoms/VisualisationParameterSelector.tsx]: I would like to further this idea by using [string enums](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums) to remove the switch statement entirely.

File: src/components/atoms/VisualisationParameterSelector.tsx [hunk 7]
```diff
@@ -19,14 +26,25 @@ const VisualisationParameterSelector: React.FC<VisualisationParameterSelectorPro
...
+    const defaultValue = () => {
+        switch (size) {
+            case DashboardItemSize.LARGE:
+                return 'large';
+            case DashboardItemSize.MEDIUM:
+                return 'medium';
+            case DashboardItemSize.SMALL:
+                return 'small';
+        }
+    };
```
FredrikAugust: See comment about string enums.

src/components/atoms/VisualisationParameterSelector.tsx
emilom: Added a new solution now which removes the switch statement.

src/components/molecules/AddToDashboard.tsx
emilom: it should work now

src/components/atoms/VisualisationParameterSelector.tsx
```diff
@@ -3,20 +3,20 @@ import { Pane, SelectField, Textarea, Label } from 'evergreen-ui';
...
     const handleSizeChange: React.ChangeEventHandler<HTMLSelectElement> = (e) => {
-        switch (e.currentTarget.value as 'small' | 'medium' | 'large') {
-            case 'small':
-                return setSize(DashboardItemSize.SMALL);
-            case 'medium':
-                return setSize(DashboardItemSize.MEDIUM);
-            case 'large':
-                return setSize(DashboardItemSize.LARGE);
-        }
+        setSize(+e.currentTarget.value);
```
FredrikAugust, line 19: If this is to convert it to a number, I'd prefer `Number(...)` as it's a lot clearer and less hacky.

FredrikAugust:
```suggestion
        setSize(Number(e.currentTarget.value));
```

FredrikAugust (approved):

PR #116, opened 2021-04-08
Present population for selected municipality

### LINKED ISSUE #112: Population data presentation
### Description
As of now, the population data visualisation doesn't work.

Your task is to figure out how it should work, and make it work.

Example solution:
- The user selects one municipality from a list of municipalities on the visualisation selection page. This requires you to perform a graphql query on the visualisation selection page to retrieve the available municipalities, as well as the years available.
- You would then need to include the selections in the global store, and map it to an input to the line/threshold charts which will then be rendered on the front page

### Subtasks
- [x] Come up with a solution
- [x] Implement said solution

fredrikbw:
Workaround using LocalStorage to use the municipality selected in the dropdown-menu.

Closes #112


src/components/atoms/MunicipalitySelector.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { SelectField } from 'evergreen-ui';
+
+type Props = {
+    label: string;
+    options: [string, string][];
+    onChange: React.ChangeEventHandler<HTMLSelectElement>;
```
FredrikAugust, line 7: Creds for getting this type right;)

src/components/atoms/MunicipalitySelector.tsx
```diff
@@ -0,0 +1,20 @@
...
+};
+
+const MunicipalitySelector: React.FC<Props> = (props) => (
+    <SelectField label={props.label} defaultValue={'Oslo'} onChange={props.onChange}>
+        {Object.entries(props.options).map((value) => (
```
FredrikAugust: Don't use Object.entries as `options: [string, string][]`. 

You can map the array directly.

Also, to avoid having to use direct indexing (`[0], [1], ...`) you can destructure the array as seen here;

```tsx
        {Object.entries(props.options).map(([key, value]) => (
```

FredrikAugust: Or in this example

- key: municipalityKey
- value: municipalityName

src/components/molecules/MunicipalityEditor.tsx
```diff
@@ -0,0 +1,44 @@
...
+        if (!data) return;
+        setselectedMunicipality(data.populationsInNorway.municipalitiesWithKeys[0][0]);
+    }, [data]);
+
+    if (error) {
+        return <Text>{error.message}</Text>;
+    }
+
+    if (loading) {
+        return <Text>Loading…</Text>;
```
FredrikAugust: Sniknorskifiser dette;)

src/components/molecules/MunicipalityEditor.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { useEffect, useState } from 'react';
+import { useQuery } from '@apollo/client';
+import { MunicipalityInNorway, MUNICIPALITY_IN_NORWAY } from '../../queries/municipalityInNorway';
+import { Text } from 'evergreen-ui';
+import MunicipalitySelector from '../atoms/MunicipalitySelector';
+
+const MunicipalityEditor: React.FC = () => {
+    const [, setselectedMunicipality] = useState('');
```
FredrikAugust: Why set it if you're not reading the value?

src/components/molecules/MunicipalityEditor.tsx
FredrikAugust: And since you have if (error), if (loading), &c. Could you use the datawrapper?

src/components/organisms/VisualisationEditor.tsx
```diff
@@ -37,6 +39,10 @@ const VisualisationEditor: React.FC = () => {
     // Safe non-null-assertion due to the .some above
     // eslint-disable-next-line
     const metadata = data.allMetadata.find((el) => el.id === id)!;
+    let municipalitySelector;
+    if (metadata.datasourceId === DataSourceID.SSB_POPULATION) {
+        municipalitySelector = <MunicipalitySelector />;
+    }
```
FredrikAugust:
```suggestion
```

src/components/organisms/VisualisationEditor.tsx
```diff
@@ -63,6 +69,7 @@ const VisualisationEditor: React.FC = () => {
                 )}
                 onChange={(e) => setSelectedVisualisation(e.currentTarget.value as VisualisationType)}
             />
+            {municipalitySelector}
```
FredrikAugust:
```suggestion
            {metadata.datasourceId === DataSourceID.SSB_POPULATION && <MunicipalitySelector />}
```

src/queries/populationInNorway.ts
```diff
@@ -1,9 +1,11 @@
 import { gql } from '@apollo/client';
 
+const code = localStorage.getItem('municipality') || 'K-0301';
+
 export const POPULATION_IN_NORWAY = gql`
     query {
         populationsInNorway {
-            dataset(years: ["2014", "2015", "2016", "2017", "2018", "2019", "2020"], municipalities: ["K-0301"]) {
+            dataset(years: ["2014", "2015", "2016", "2017", "2018", "2019", "2020"], municipalities: ["${code}"]) {
```
FredrikAugust: See docs about parameters:

https://www.apollographql.com/docs/react/data/queries/#usequery-api

src/components/organisms/VisualisationEditor.tsx
fredrikbw: Thanks

src/components/organisms/VisualisationEditor.tsx
```diff
@@ -72,6 +75,12 @@ const VisualisationEditor: React.FC = () => {
                 )}
                 onChange={(e) => setSelectedVisualisation(e.currentTarget.value as VisualisationType)}
             />
+            {metadata.datasourceId === DataSourceID.SSB_POPULATION && (
+                <>
+                    <MunicipalitySelector />
+                    <YearSelector />
+                </>
+            )}
```
FredrikAugust: I think we should talk about how to solve/scale this.

src/queries/populationInNorway.ts
```diff
@@ -1,9 +1,17 @@
...
+const years = new Array<string>();
+for (let index = Number(beginYear); index <= Number(endYear); index++) {
+    years.push(index.toString());
+}
+
 export const POPULATION_IN_NORWAY = gql`
     query {
         populationsInNorway {
-            dataset(years: ["2014", "2015", "2016", "2017", "2018", "2019", "2020"], municipalities: ["K-0301"]) {
+            dataset(years: [${years.map((x) => `"` + x + `",`)}], municipalities: ["${code}"]) {
```
FredrikAugust: Use variables.

src/queries/populationInNorway.ts
```diff
@@ -1,9 +1,17 @@
 import { gql } from '@apollo/client';
 
+const code = localStorage.getItem('municipality') || 'K-0301';
+const beginYear = localStorage.getItem('beginYear') || '1986';
+const endYear = localStorage.getItem('endYear') || '2020';
+const years = new Array<string>();
+for (let index = Number(beginYear); index <= Number(endYear); index++) {
+    years.push(index.toString());
+}
```
FredrikAugust: Try to avoid for-loops where possible

src/queries/populationInNorway.ts
```diff
@@ -1,9 +1,17 @@
 import { gql } from '@apollo/client';
 
+const code = localStorage.getItem('municipality') || 'K-0301';
+const beginYear = localStorage.getItem('beginYear') || '1986';
+const endYear = localStorage.getItem('endYear') || '2020';
+const years = new Array<string>();
```
FredrikAugust:
```suggestion
const years: string[] = [];
```

src/queries/populationInNorway.ts
```diff
@@ -1,9 +1,12 @@
 import { gql } from '@apollo/client';
+import { store } from '../redux/index';
+
+const municipality = store.getState().municipality.selected;
 
 export const POPULATION_IN_NORWAY = gql`
     query {
         populationsInNorway {
-            dataset(years: ["2014", "2015", "2016", "2017", "2018", "2019", "2020"], municipalities: ["K-0301"]) {
+            dataset(years: ["2014", "2015", "2016", "2017", "2018", "2019", "2020"], municipalities: ["${municipality}"]) {
```
FredrikAugust, line 9: Use GraphQL parameters

src/redux/slices/municipality.ts
```diff
@@ -0,0 +1,27 @@
...
+    initialState,
+    reducers: {
+        add: (state, action: PayloadAction<string>) => {
+            state.selected = action.payload;
+        },
+        update: (state, action: PayloadAction<string>) => {
+            state.selected = action.payload;
+        },
+        remove: (state) => {
+            if (state) {
```
FredrikAugust, line 22: How can state be nullable?

FredrikAugust:
I don't think you need to introduce a reducer for this. 

You should be able to extend the `DashboardVisualisation` type to include some `DataSourceOptions` that can generate controls on the `VisualisationEditor` page. That way it could also be stored in `LocalStorage`, and it will be unique for _one_ visualisation, which is what we want.

PR #118, opened 2021-04-09
Fix firefox visualisation preview bug

### LINKED ISSUE #110: Small/Medium visualisation doesn't look healthy on FF
### Description
Could be an idea to rethink the sizing implementation.

### Steps to reproduce
1. Use firefox
2. Go to vis editor
3. Select small/medium

### Expected behaviour
Correct sizing and positioning

### Details
 - OS: MacOS
 - Browser FF
 - Version 87

### Additional comments
Screenshot 2021-04-07 at 14 58 57

To me it's unnatural that it is pushed to the right on safari/chrome.

emilom:
Changed how the visualisation preview css works by using flex and % instead of a hacky grid solution. This should work on firefox, chrome and safari now.  I don't like how the "Legg til i dashboard" button is placed when the visualisation is medium or small. Do you @thomran have any suggestions on how to fix this?  

Close #110

### DIFF
--- a/src/components/molecules/VisualisationPreview.tsx
+++ b/src/components/molecules/VisualisationPreview.tsx
@@ -81,13 +81,20 @@ const VisualisationPreview: React.FC<VisualisationPreviewProps> = ({ metadata, s
         return null;
     }
 
+    const getSizeInPercentage = () => {
+        switch (size) {
+            case DashboardItemSize.LARGE:
+                return '100%';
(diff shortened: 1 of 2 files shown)


emilom:
And the page looks a bit ugly when a visualisation has tons of tags. Should we cap the amount of tags displayed here?

FredrikAugust (approved):

PR #119, opened 2021-04-09
Set selected visualisation if added to global store

emilom:
Saves the selected visualisation when you refresh the visualisation editor page. Will also save the visualisation if you go to an visualisation in the data explorer you have already added.

### DIFF
--- a/src/components/atoms/VisualisationSelector.tsx
+++ b/src/components/atoms/VisualisationSelector.tsx
@@ -1,14 +1,16 @@
 import React from 'react';
 import { SelectField } from 'evergreen-ui';
+import { VisualisationType } from '../../types/Metadata';
 
 type Props = {
+    default: VisualisationType;
     label: string;
     options: Record<string, string>;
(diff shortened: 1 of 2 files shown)


src/components/atoms/VisualisationSelector.tsx
```diff
@@ -1,14 +1,16 @@
...
 type Props = {
+    default: VisualisationType;
     label: string;
     options: Record<string, string>;
     onChange: React.ChangeEventHandler<HTMLSelectElement>;
 };
 
 const VisualisationSelector: React.FC<Props> = (props) => (
-    <SelectField label={props.label} defaultValue={Object.values(props.options)[0]} onChange={props.onChange}>
+    <SelectField label={props.label} value={props.default} onChange={props.onChange}>
```
FredrikAugust, line 13: Why did you remove defaultValue?

emilom: The default value would only work if you re-freshed the page or visited it for the first time.

FredrikAugust (approved):

PR #122, opened 2021-04-14
Search and filter datasets by tags

### LINKED ISSUE #132: Allow searching for data sets
### Description
Allow the user to search for a data set by the title of the DS.

### Subtasks
- [x] Create UI
- [x] Implement search functionality (frontend only)

emilom:
Ui and functionality for searching and filtering datasets by tags. 
This is done with a fixed sidebar, where all tags are fetched from the cache and displayed. 
All selected tags are stored with redux and the filtering is done in the frontend in the DataResultItems.tsx organism. 
Close #132

Never used redux before, so feedback is appreciated <3


src/components/atoms/SelectTag.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+type SelectTagProps = {
+    addTag: (tag: string) => void;
+};
+
+const SelectTag: React.FC<SelectTagProps> = ({ addTag }) => {
+    const { data } = useQuery<AllTags>(TAGS);
+
+    const tags = data?.allMetadata.map((e) => e.tags).flat(1);
+    if (!tags) return <p>Rip, ingen tags</p>;
```
FredrikAugust: Maybe a bit more formal language

FredrikAugust: Why aren't `loading` and `error` handled?

src/components/organisms/DataResultItems.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { useSelector } from 'react-redux';
+import { RootState } from '../../redux';
+
+const DataResultItems: React.FC = () => {
+    const { data, loading, error } = useQuery<AllMetadataResult>(METADATA);
+    const activeFilters = useSelector((state: RootState) => state.filter.filters);
+    const result = useMemo(() => {
+        if (!data) return null;
+        if (activeFilters.length === 0) return data.allMetadata;
+        return data?.allMetadata.filter((el) => el.tags.some((v) => activeFilters.indexOf(v) !== -1));
```
FredrikAugust: Use [`includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes).

FredrikAugust: How can data be null here? Isn't it checked two lines above?

src/components/organisms/DataResultItems.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { METADATA, AllMetadataResult } from '../../queries/metadata';
+import Loading from '../atoms/Loading';
+import ErrorMessage from '../atoms/ErrorMessage';
+import { Link } from 'react-router-dom';
+import { useSelector } from 'react-redux';
+import { RootState } from '../../redux';
+
+const DataResultItems: React.FC = () => {
+    const { data, loading, error } = useQuery<AllMetadataResult>(METADATA);
+    const activeFilters = useSelector((state: RootState) => state.filter.filters);
```
FredrikAugust: Since the filters are entirely limited to one page, I don't know if Redux is strictly necessary, or if you could just create a simple reducer with `useReducer` in this component.

That way you would limit the state to the relevant component.

FredrikAugust: Since you're also just dealing with a list of tags, it might even be solvable with only a `useState`.

When updating a `useState` hook, you can write something like this;

```ts
setState((previousState) => ([...previousState, newTag]))
```

To add a new entry to the list.

FredrikAugust: Though a reducer might open for more future functionality. 

Another benefit of using a `useReducer` here instead of redux is that you eliminate a dependency from the component subtree (i.e. you no longer require a provider or library).

FredrikAugust (changes requested):
Misclick. Meant to request changes.

FredrikAugust:
Could you also add some storybook entries for this with the different states?

src/components/atoms/ActiveFilters.tsx
```diff
@@ -0,0 +1,32 @@
...
+    tags: string[];
+    removeTag: (tag: string) => void;
+    removeAll: () => void;
+};
+
+const ActiveFilters: React.FC<ActiveFiltersProps> = ({ tags, removeTag, removeAll }) => {
+    return (
+        <Pane marginBottom="1rem">
+            <Pane display="flex" justifyContent="space-between" alignItems="center">
+                <Text size={300}>AKTIVE FILTERE</Text>
```
FredrikAugust: https://ordbok.uib.no/perl/ordbok.cgi?OPP=filter&ant_bokmaal=5&ant_nynorsk=5&begge=+&ordbok=begge

src/components/atoms/ActiveFilters.tsx
```diff
@@ -0,0 +1,32 @@
...
+                <Button appearance="minimal" intent="none" onClick={removeAll}>
+                    Nullstill
+                </Button>
+            </Pane>
+            <Card border="default" display="flex" flexWrap="wrap" minHeight="4rem" padding="0.2rem" background="white">
+                {tags.length === 0 && <Text color="muted">Ingen filtere valgt</Text>}
```
FredrikAugust: https://ordbok.uib.no/perl/ordbok.cgi?OPP=filter&ant_bokmaal=5&ant_nynorsk=5&begge=+&ordbok=begge

FredrikAugust (changes requested):
Looks a lot better!

Some changes to be closer to the design specification;
- The filter panel could probably be a bit larger, right now it's quite thin
- There should be more padding around the controls
- The "Nullstill" button should be disabled when there are no tags
- There shouldn't be a "Ingen filtere valgt" label when no labels are selected

src/components/atoms/ActiveFilters.tsx
emilom: @thomran

FredrikAugust (approved):

PR #135, opened 2021-04-14
Add toaster messages on add/delete events

### LINKED ISSUE #130: Toast important events using Evergreen toaster
### Description
We should let the user know when important events happen

### Subtasks
- [x] Create visualisation
- [ ] Delete visualisation

### Related sources
- [Documentation](https://renaafjellet.no/wp-content/uploads/2020/09/Ost-og-skinke-toast.jpg)

Kpaubert:
Closes #130 
Added toaster messages on add/delete visualisations.

### DIFF
--- a/src/components/molecules/AddToDashboard.tsx
+++ b/src/components/molecules/AddToDashboard.tsx
@@ -22,13 +22,13 @@ const AddToDashboard: React.FC<Props> = (props) => {
 
     const add = () => {
         dispatch(dashboard.actions.add(props.dashboardItem));
-        toaster.success('Visualiseringen ble lagt til i dashboardet ditt.');
+        toaster.success('Visualiseringen ble lagt til i dashboardet.');
         history.push('/');
     };
 
     const remove = () => {
         dispatch(dashboard.actions.remove(key));
-        toaster.success('Visualiseringen ble fjernet.');
+        toaster.success('Visualiseringen ble fjernet fra dashboardet.');
     };
 
     return <AddToDashboardButton added={added} onAdd={add} onRemove={remove} />;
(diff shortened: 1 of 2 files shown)


src/components/molecules/AddToDashboard.tsx
```diff
@@ -21,11 +22,13 @@ const AddToDashboard: React.FC<Props> = (props) => {
 
     const add = () => {
         dispatch(dashboard.actions.add(props.dashboardItem));
+        toaster.success('Visualiseringen ble lagt til i dashboardet.');
         history.push('/');
     };
 
     const remove = () => {
         dispatch(dashboard.actions.remove(key));
+        toaster.success('Visualiseringen ble fjernet fra dashboardet.');
```
FredrikAugust, line 31: Is deleting a `success` action? I think it may look more nuanced with `info`.

Please share your thoughts.

FredrikAugust (commented):
Should the toast messages end with a full stop?

Kpaubert:
Not sure about the full stops , but it probably looks better without them :)

src/components/molecules/AddToDashboard.tsx
Kpaubert: I think success is the most fitting for both adding and removing, since it conveys to the user that it succeeded in doing what the user asked. Having consistent feedback for these operations (adding/deleting) is good, since these are the operations in which toasts are used. _These are (naturally) my subjective thoughts._

How _I_ think the logic of toasters should be:
* `success`: any user operation that was successful (e.g. "Successfully deleted the entire database")
* `notify`/`info`: useful information not necessarily related to user interacting with the page (e.g. "Check out this offer")
* `warning`: trivial user mistakes such as not filling in a form (e.g. "Please fill in your email") 
* `error`:  errors rooted in a user action that failed (not necessarily caused by the user, could be a server error: "Could not update username: `${error.message}`")

Let me know if this is totally stupid 🚀

FredrikAugust: Good reflections! I usually use notify and info for more user-interactive actions, but consistency is the most important.

FredrikAugust (approved):

PR #136, opened 2021-04-15
Replace visx with plotly

### LINKED ISSUE #120: Replace custom visualisations with graphs from decided backup graph library
### Description
As discussed in the meeting, we should use the pre-made charts instead as they are _better_.

### Subtasks
- [x] Find out which library it is
- [x] Replace LineChart
- [x] Replace ThresholdChart (if possible)

### LINKED ISSUE #128: White background colour on smaller dashboard items
This leaves a white background on size medium or small. I think the white background should be set in dashboarditem instead of the container.

_Originally posted by @emilom in https://github.com/IT2901-SINTEF01/frontend/pull/124#discussion_r613387735_

### LINKED ISSUE #133: Show added text in visualisation preview
Adding text will not show up in preview when using visualisation editor

emilom:
Changed visx visualisations to plotly. 

TODO
- [x] Make sure NPM 📦 works.
- [x] Update storybook to work with redux and fix all controls.
- [x] Some bugs with scaling, changing size in preview and adding text. 

Close #120 
Close #133
Close #128


FredrikAugust:
Does storybook work now?

src/components/molecules/DataResultItem.tsx
```diff
@@ -17,6 +16,35 @@ export type DataResultItemProps = {
...
+                        yLabel="y"
+                        isPreview={true}
+                        height="100%"
+                        width="100%"
+                    />
+                );
+            default:
+                return <Text>{visualisationType}</Text>;
+        }
+    };
```
FredrikAugust: Could we move this to a dedicated `PreviewVisualisation.tsx` file and component?

src/components/molecules/VisualisationPreview.tsx
```diff
@@ -101,11 +105,7 @@ const VisualisationPreview: React.FC<VisualisationPreviewProps> = ({
                         titleSize={400}
                         paragraph={paragraph}
                     >
-                        <ParentSize debounceTime={400}>
-                            {(parent) => {
-                                return child(parent.width, parent.height);
-                            }}
-                        </ParentSize>
+                        <ParentSize>{(parent) => child(parent.width)}</ParentSize>
```
FredrikAugust: Do we need this? Can't it be set to 100% width?

src/components/molecules/VisualisationPreview.tsx
```diff
@@ -33,28 +33,32 @@ const VisualisationPreview: React.FC<VisualisationPreviewProps> = ({
             return () => null;
         }
 
-        const timeEntryMockData = mockTimeEntry(100, visualisation.axes.y.limit);
+        const timeEntryMockData = mockTimeEntry(
+            100,
+            selectedVisualisation === VisualisationType.THRESHOLD ? [-100, 100] : visualisation.axes.y.limit,
+        );
 
         switch (visualisation.type) {
```
FredrikAugust: Could this also be moved to a dedicated file and component?

src/components/molecules/VisualisationPreview.tsx
```diff
@@ -33,28 +33,32 @@ const VisualisationPreview: React.FC<VisualisationPreviewProps> = ({
             return () => null;
         }
 
-        const timeEntryMockData = mockTimeEntry(100, visualisation.axes.y.limit);
+        const timeEntryMockData = mockTimeEntry(
+            100,
+            selectedVisualisation === VisualisationType.THRESHOLD ? [-100, 100] : visualisation.axes.y.limit,
+        );
```
FredrikAugust: Could you extend `mockTimeEntry` to take in the type it should generate data for? So we can avoid ternary expressions in function parameters.

```ts
function (numberOfEntries: number, visualisationType: VisualisationType) {
  // ...
}
```

src/components/visualisations/LineChart.tsx
```diff
@@ -1,91 +1,59 @@
...
+                    r: 40,
+                    b: 40,
+                    t: 40,
+                    pad: 4,
+                },
+                yaxis: {
+                    title: yLabel,
+                },
+            }}
+            config={{ staticPlot: isPreview }}
```
FredrikAugust, line 56: Cool :o

FredrikAugust (changes requested):

FredrikAugust:
Screenshot 2021-04-15 at 13 05 06



The thresholdchart is rather difficult to read. Is it possible to move the X-axis line up to the threshold?

src/components/molecules/DataResultItem.tsx
```diff
@@ -17,6 +16,35 @@ export type DataResultItemProps = {
...
     const timeEntryMockData = mockTimeEntry(100);
 
+    const visualisation = () => {
+        switch (visualisationType) {
+            case VisualisationType.LINE:
+                return (
+                    <LineChart
+                        data={timeEntryMockData}
+                        strokeColor="#66CCCC"
+                        yLabel="y"
```
FredrikAugust: This looks rather strange. Could we perhaps omit it?

src/components/molecules/DataResultItem.tsx
emilom: yes!

src/components/molecules/VisualisationPreview.tsx
FredrikAugust: Could this be merged with the extracted switch in `DataResultItem`? So we could [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) up the code some

src/components/molecules/VisualisationPreview.tsx
emilom: Unfortunately plotly doesn’t redraw the graph unless the styling or data changes, or the window is resized. You could probably force a re render somehow but this was the easiest fix i could find.

src/components/molecules/DataResultItem.tsx
emilom: the label?

src/components/molecules/VisualisationPreview.tsx
emilom: > Does storybook work now?

Should work now. Had to wrap storybook preview.jsx in a redux provider.

src/components/molecules/DataResultItem.tsx
FredrikAugust: Yes

FredrikAugust:
Screenshot 2021-04-15 at 13 15 52


The Y-label floats into the axis ticks.

emilom:
> Screenshot 2021-04-15 at 13 05 06
> 
> The thresholdchart is rather difficult to read. Is it possible to move the X-axis line up to the threshold?

Yeah thats a mistake. The X-axis line should be at the threshold value.

blauks (changes requested):
Nice, looks good! But before we merge this we should uninstall all the unused visx packages. Also, can we change the name of ```mockTimeEntry``` to ```mockCartesianChartInput``` (or something) now,  since the TimeEntry type is gone 😀

src/components/molecules/VisualisationPreview.tsx
blauks: But won't the component rerender when you change the ```visualisation``` prop?

emilom: yup, but not when we are only changing the size of the parent

emilom: nvm, i found a fix 👍

blauks (dismissed):
lgtm

FredrikAugust:
Going to re-review later

src/components/molecules/MockedVisualisation.tsx
```diff
@@ -0,0 +1,32 @@
...
+    height: string | number;
+    width: string | number;
+};
+
+const MockedVisualisation: React.FC<MockedVisualisationProps> = ({ visualisationType, height, width }) => {
+    const [data, setData] = useState<CartesianChartInput>(mockCartesianChartInput(100, visualisationType));
+
+    useEffect(() => {
+        setData(mockCartesianChartInput(100, visualisationType));
+    }, [visualisationType]);
```
FredrikAugust: What is this for?

```ts
const data = mockCartesianChartInput(100, visualisationType);
```

Wouldn't this do the same?

src/components/molecules/VisualisationPreview.tsx
```diff
@@ -93,19 +48,19 @@ const VisualisationPreview: React.FC<VisualisationPreviewProps> = ({
                     </Heading>
                     <Pane flex="1" />
                 </Pane>
-                <Pane backgroundColor="white">
+                <Pane>
```
FredrikAugust: Do you need this Pane?

FredrikAugust (changes requested):

src/components/molecules/MockedVisualisation.tsx
FredrikAugust: When `visualisationType` changes `data` will be recalculated anyway due to prop changes.

Try to avoid using `useEffect`s where they aren't needed. It might help to try _not_ using them before resorting to them – and you might avoid some review questions while you're at it;)

FredrikAugust: You also don't need to specify the type of the `useState` when you're initialising the state immediately. It's inferred.

emilom: It would create a new graph each time the window is resized or if the visualisation is changed to a different size

FredrikAugust: I see.

I would recommend looking into the available hooks in React, and perhaps particularly `useMemo`.

FredrikAugust: The pattern of `useState` plus a `useEffect` is rather convoluted and difficult to understand right away, and requires duplicating code which is something we should aim to avoid.

src/components/molecules/VisualisationPreview.tsx
emilom: Nope

FredrikAugust (approved):

PR #143, opened 2021-04-21
Make 'legg til i ditt dashboard' button scale with different viewport sizes

### LINKED ISSUE #127: "Legg til i ditt dashboard"  button doesn't fit on smaller screen sizes
The text of this button does not fit on my screen. Maybe remove some text or make the button higher?

_Originally posted by @emilom in https://github.com/IT2901-SINTEF01/frontend/pull/124#discussion_r613383840_

emilom:
The button now looks good on all viewport sizes.

Close #127

### DIFF
--- a/cypress/integration/website.spec.tsx
+++ b/cypress/integration/website.spec.tsx
@@ -34,7 +34,7 @@ describe('Dashboard', () => {
         cy.visit('/');
         cy.contains('Legg til nytt komponent').click();
         cy.wait('@METADATA');
-        cy.get('a').first().click();
+        cy.get('.dataResultItems').first().click();
         cy.contains('Legg til i dashboard').click();
         cy.get('.dashboardItem').should('be.visible');
     });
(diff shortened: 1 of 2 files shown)


FredrikAugust (changes requested):
The new design:
Screenshot 2021-04-21 at 10 47 42

The old design:
Screenshot 2021-04-21 at 10 48 07

-------

The new design is no longer in thread with the Evergreen UI component design, and thus not in alignment with the design from @thomran.

I believe it would be beneficial if we kept a uniform set of UI components.

Other approaches to this;
- `line-height: 1em` on the button/text
- Make the control size bigger
- Confer Thomas about moving the button

emilom:
The new design:
Screenshot 2021-04-21 at 10 47 42

The updated new design:
Screenshot 2021-04-21 at 12 20 16

The old design:
Screenshot 2021-04-21 at 10 48 07

-------


The font size is still off and differs from other buttons in the library.

Could you perhaps just shorten the text in the button, or set the line height to e.g. `1.2em` instead? Do we need the entire "Legg til i **ditt** dashboard"? Could we omit "ditt"?

This is of course just nit-pick, but differing from the component library is something best avoided unless absolutely needed.

This would also serve as something worth mentioning in the report; Evergreen isn't a very mature library and is missing some features we would benefit from.

FredrikAugust:
And if not, could you perhaps try inherting font size (and eventual other differing attributes) from the button? 

https://developer.mozilla.org/en-US/docs/Web/CSS/inherit

FredrikAugust (changes requested):
See comments above.

FredrikAugust:
LGTM as soon as the tests are updated!

FredrikAugust (dismissed):

FredrikAugust (approved):
GJ!

PR #144, opened 2021-04-21
Update package version

blauks:
This PR updates the NPM package version to 2.0.0 since we have new graphs, with different props. Package has also been published to NPM:)


emilom (approved):
LVGTM

PR #145, opened 2021-04-21
139 allow multiple instances of same vis

Kpaubert:
Change method of locally storing visualisations to include a UUID, as well as the metadata id. There are some problems when using this with mocked data, as the UUID's sometimes change, but this should not be a problem in prod.

*Test thoroughly before merging*


src/components/organisms/DashboardItems.tsx
```diff
@@ -6,7 +6,7 @@ import { ParentSize } from '@visx/responsive';
 import DataWrapper from '../molecules/DataWrapper';
 import { useDispatch, useSelector } from 'react-redux';
 import { RootState } from '../../redux';
-import visualisationMapping, { VisualisationMappingFunctionPath } from '../../utils/visualisationMapping';
+// import visualisationMapping, { VisualisationMappingFunctionPath } from '../../utils/visualisationMapping';
```
blauks: Hey, it looks like you have commented this out? Are there any other ways to remove this import?

src/utils/visualisationMapping.ts
```diff
@@ -1,10 +1,12 @@
-import { DataSourceID } from '../types/DataSource';
-import { VisualisationType } from '../types/Metadata';
+// import { DataSourceID } from '../types/DataSource';
+// import { VisualisationType } from '../types/Metadata';
```
blauks: Hey again

FredrikAugust:
Relevant lines from failed build

```
#17 65.41 Could not find a declaration file for module 'uuid'. '/app/node_modules/uuid/dist/index.js' implicitly has an 'any' type.
#17 65.41   Try `npm i --save-dev @types/uuid` if it exists or add a new declaration (.d.ts) file containing `declare module 'uuid';`  TS7016
```

src/components/molecules/AddToDashboard.tsx
```diff
@@ -15,7 +15,7 @@ type Props = {
 const AddToDashboard: React.FC<Props> = (props) => {
     const history = useHistory();
 
-    const key = `${props.dashboardItem.dataSourceId}-${props.dashboardItem.visualisationType}` as VisualisationMappingFunctionPath;
+    const key = props.dashboardItem.id ?? uuidv4();
```
FredrikAugust: I think it might be beneficial if this doesn't change when dashboardItem changes, and only when it's (re)mounted.

You can solve this with a `useMemo`.

src/components/organisms/DashboardItems.tsx
FredrikAugust: Refrain from using comments in production-ready code.

src/components/organisms/DashboardItems.tsx
```diff
@@ -65,14 +66,16 @@ const DashboardItems: React.FC = () => {
                             width="100%"
                             titleSize={400}
                             paragraph={visualisation.options.paragraph}
-                            onEdit={() => history.push(`/explore/edit/${metadataEntry?.id}`)}
+                            onEdit={() => history.push(`/explore/edit/${mappingPath}`)}
```
FredrikAugust: If `mappingPath` is now the UUID, update the name accordingly.

src/components/pages/VisualisationEditor.tsx
```diff
@@ -33,15 +37,19 @@ const VisualisationEditor: React.FC = () => {
         return <Text>Loading…</Text>;
     }
 
-    if (!data || !data.allMetadata.some((el) => el.id === id)) {
+    if (typeof data === 'undefined' || ((!data || !data.allMetadata.some((el) => el.id === id)) && !visualisation)) {
```
FredrikAugust: Why is `typeof data === 'undefined'` used instead of just `!data` or `data === undefined`?

blauks (dismissed):
Hey

src/components/pages/VisualisationEditor.tsx
FredrikAugust: This `if`-statement would also benefit from being broken up into smaller, more legible pieces. Right now it's a bit difficult to understand.

src/components/pages/VisualisationEditor.tsx
```diff
@@ -33,15 +37,19 @@ const VisualisationEditor: React.FC = () => {
...
     }
 
-    // Safe non-null-assertion due to the .some above
-    // eslint-disable-next-line
-    const metadata = data.allMetadata.find((el) => el.id === id)!;
+    const metadata = visualisation
+        ? data.allMetadata.find((el) => el.id === visualisation.metadataId)
+        : data.allMetadata.find((el) => el.id === id);
+
+    if (typeof metadata === 'undefined') {
```
FredrikAugust: Again, use `!metadata` or `metadata === undefined`

src/redux/slices/dashboard.ts
```diff
@@ -1,22 +1,29 @@
...
 
 export default createSlice({
     name: 'Dashboard Items',
-    initialState: {} as Record<VisualisationMappingFunctionPath, DashboardVisualisation>,
+    initialState: {} as Record<string, DashboardVisualisation>,
     reducers: {
         add: (state, action: PayloadAction<DashboardVisualisation>) => {
-            const key = `${action.payload.dataSourceId}-${action.payload.visualisationType}` as VisualisationMappingFunctionPath;
-            state[key] = action.payload;
+            const key = `${uuidv4()}`;
```
FredrikAugust: If it's a string, use `uuidv4()`, and if it can be casted to it, use `uuidv4() as string`, and if should be converted, use `String(uuidv4)`. These are all more explicit (or in the former case, implicit) and should be preferred to templates.

src/redux/slices/dashboard.ts
```diff
@@ -1,22 +1,29 @@
...
+            state[key] = {
+                id: key,
+                dataSourceId: action.payload.dataSourceId,
+                visualisationType: action.payload.visualisationType,
+                options: action.payload.options,
+                metadataId: action.payload.metadataId,
```
FredrikAugust:
```suggestion
                ...action.payload,
```

src/redux/slices/dashboard.ts
```diff
@@ -1,22 +1,29 @@
...
-                delete state[action.payload];
+        remove: (state, action: PayloadAction<string>) => {
+            const key = `${action.payload}`;
+            if (state[key]) {
+                delete state[key];
             }
         },
         update: (state, action: PayloadAction<DashboardVisualisation>) => {
-            const key = `${action.payload.dataSourceId}-${action.payload.visualisationType}` as VisualisationMappingFunctionPath;
+            const key = `${action.payload.id}`;
```
FredrikAugust: Again, no need for templates

src/redux/slices/dashboard.ts
```diff
@@ -1,22 +1,29 @@
...
+                visualisationType: action.payload.visualisationType,
+                options: action.payload.options,
+                metadataId: action.payload.metadataId,
+            };
         },
-        remove: (state, action: PayloadAction<VisualisationMappingFunctionPath>) => {
-            if (state[action.payload]) {
-                delete state[action.payload];
+        remove: (state, action: PayloadAction<string>) => {
+            const key = `${action.payload}`;
```
FredrikAugust: :cat2:

src/utils/visualisationMapping.ts
FredrikAugust: No comments

src/types/DashboardVisualisation.ts
```diff
@@ -26,7 +26,9 @@ export enum DashboardItemSize {
 }
 
 export type DashboardVisualisation = {
+    id?: string;
```
FredrikAugust: Is this optional?

FredrikAugust (changes requested):

src/components/pages/VisualisationEditor.tsx
Kpaubert: Because typescript does not accept it 🗡️

Kpaubert: image

FredrikAugust: Then `data === undefined` would be a cleaner check.

Generally, using `typeof xxx === ` should be avoided unless you're working with custom classes and need to check for instances of them.

FredrikAugust: Typeof is rather strange with quirks like

```
// This stands since the beginning of JavaScript
typeof null === 'object';
```

FredrikAugust: If it cannot be null, then an explicit check for `data === undefined` should suffice.

src/types/DashboardVisualisation.ts
Kpaubert: Yes :D

FredrikAugust: The Visualisation on the dashboard needs to have an `id`. Is it possible to rewrite the code such that this type can have a `NonNullable` `id`?

FredrikAugust: Perhaps by immediately assigning an `id` field in the cases where it's not present. I.e. move it from the `AddToDashboard` atom to somewhere else

FredrikAugust: I think it would be cleaner if the Dashboard didn't have to care about _how_ the objects are before they end up on the dashboard.

src/types/DashboardVisualisation.ts
```diff
@@ -26,7 +26,9 @@ export enum DashboardItemSize {
...
     dataSourceId: DataSourceID;
     visualisationType: VisualisationType;
     options: ChartCustomisationOption;
+    metadataId: string;
```
FredrikAugust: What is the purpose of this ID? Can't this be inferred from dataSourceId since each metadata is one-to-one with data sources?

src/redux/slices/dashboard.ts
```diff
@@ -1,22 +1,25 @@
...
-    initialState: {} as Record<VisualisationMappingFunctionPath, DashboardVisualisation>,
+    initialState: {} as Record<string, DashboardVisualisation>,
     reducers: {
         add: (state, action: PayloadAction<DashboardVisualisation>) => {
-            const key = `${action.payload.dataSourceId}-${action.payload.visualisationType}` as VisualisationMappingFunctionPath;
-            state[key] = action.payload;
+            const key = uuidv4();
+            state[key] = {
+                ...action.payload,
+            };
```
FredrikAugust:
```suggestion
            state[key] = action.payload;
```

src/components/pages/VisualisationEditor.tsx
```diff
@@ -33,15 +37,19 @@ const VisualisationEditor: React.FC = () => {
...
         // No metadata was found. Go back
         history.push('/explore');
         return null;
     }
 
-    // Safe non-null-assertion due to the .some above
-    // eslint-disable-next-line
-    const metadata = data.allMetadata.find((el) => el.id === id)!;
+    const metadata = visualisation
+        ? data.allMetadata.find((el) => el.id === visualisation.metadataId)
```
FredrikAugust: Could you compare this with `el.dataSourceId` and similar on visualisation instea

FredrikAugust (changes requested):

src/types/DashboardVisualisation.ts
Kpaubert: Trying to fix this =)

src/types/DashboardVisualisation.ts
Kpaubert: Changing to use `dataSourceId`

src/components/pages/VisualisationEditor.tsx
Kpaubert: Yes sir

src/types/DashboardVisualisation.ts
Kpaubert: Ah nevermind. The reason I added this is because when you click `edit` on a visualization from the Dashboard, you land in something like `.../explore/edit/{uuid for the given visualization}`. If you then delete the visualization, you are redirected to `.../explore/edit/{metadata id}`, since the UUID for the given visualization does not exist any longer.

FredrikAugust: The parent of the add to dashboard button has the metadata entry, so you can pass that down to the button, and thus _should_ be able to avoid the field in the DashboardVisualisation

Kpaubert: Will fix this later

FredrikAugust (dismissed):

FredrikAugust (approved):

Issue #131, opened 2021-04-14
Support updating visualisations


FredrikAugust:
### Description
Allow the user to update visualisations

### Subtasks
- [ ] #126 
- [ ] Persist change

Issue #126, opened 2021-04-14
Hydrate edit page with parameters from visualisation


FredrikAugust:
### Description
When you click "edit" the parameters should be filled out.

### Subtasks
- [ ] `paragraph`
- [ ] `size`
- [ ] Eventual other parameters

PR #146, opened 2021-04-21
🪂  Allow user to enter data source variables on visualisationeditor page

### LINKED ISSUE #141: Allow arbitrary data source options in DashboardVisulisation
### Description
Bear with me, this might get a bit confusing. If so don't hesitate to ask here, send me a DM, or call me.

In the current drafted implementation of population parameters, i.e.

https://github.com/IT2901-SINTEF01/frontend/pull/116/files#diff-b01a6c1c95562d08ab752970c643e4069e70de83e390189d9bbd8f224a18fe21R100

There's essentially a switch statement that generates a set of controls for the data source in-place. This data is then added to a _global_ store.

https://github.com/IT2901-SINTEF01/frontend/pull/116/files#diff-7eeb8bab2802ad9120bb9d23ca0003af1c535416e48e97f300fa83fea064da9aR11-R12

https://github.com/IT2901-SINTEF01/frontend/pull/116/files#diff-7cd1314945ef5b0ae105c7554edbd674636614af44d9ba11b8e6f01b1e906909R26-R28

Firstly, we will address the state management. It is beneficial to the complexity of the application if state is stored at the [Lowest Common Ancestor](https://en.wikipedia.org/wiki/Lowest_common_ancestor) (apologies if you suffer from Post-AlgDat Traumas). This is because this isolates the state and ensures _only_ the relevant parts of the application get access to the state.

This LCA will probably be the `VisualisationEditor`, and thus state should be kept there.

---------

The second part revolves around the creation of the components to control the inputs --- and the subsequent state modifications.

_What if_, we add a field to the `DashboardVisualisation` interface called `DataSourceOptions` which has the following form (using [type aliases](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-aliases));

```ts
type GraphQLParameterName = string;
type ControlType = "municipality_editor" | "lat_lon_editor";
type ControlValue = string | number | string[]; // e.g.

type DataSourceOption = { controlType: ControlType, value: ControlValue };

type DataSourceOptions = Record<GraphQLParameterName, DataSourceOption>;
```

> The code is meant as inspiration and should be taken with a pinch of salt.

Then we could add a section in the visualisation editor which essentially iterates through the entries in the datasourceoptions and generates the relevant components. The results from these components should be injected into the GraphQL queries as parameters, and stored in the Redux store (as they should be persisted so the variables are the same on the next load).

### Subtasks
- [ ] Move state to LCA
- [ ] Implement generic control component "generator" and associated types

### Related sources
- https://github.com/IT2901-SINTEF01/frontend/blob/master/src/types/DashboardVisualisation.ts#L31

FredrikAugust:
Closes #116 
Closes #141 

--------
Stole the municipality logic from #116 (@fredrikbw).

On the visualisation editor page you can now see options for each data source and customise these to your liking. These parameters will then be included in the queries to the backend.

### DIFF
--- a/src/components/atoms/LatLonInputSet.tsx
+++ b/src/components/atoms/LatLonInputSet.tsx
@@ -1,0 +1,29 @@
+import React, { ChangeEvent, Dispatch, SetStateAction, useEffect, useState } from 'react';
+import { MetAPIVariables } from '../../queries/metApi';
+import { TextInputField } from 'evergreen-ui';
+
+type LatLonInputSetProps = {
+    state: MetAPIVariables;
+    setState: Dispatch<SetStateAction<MetAPIVariables>>;
+};
+
+const LatLonInputSet: React.FC<LatLonInputSetProps> = (props) => {
+    const [state, setState] = useState<{ lat: string; lon: string }>({ lat: '63', lon: '11' });
+
+    // To avoid problems with difficult digit entry for user
+    useEffect(() => {
+        props.setState({ lat: Number(state.lat), lon: Number(state.lon) });
+    }, [state]);
+
+    const update = (key: keyof typeof state) => (e: ChangeEvent<HTMLInputElement>) =>
+        setState((prevState) => ({ ...prevState, [key]: e.target.value }));
+
+    return (
+        <>
+            <TextInputField value={state.lat} label="Lengdegrad" onChange={update('lat')} />
+            <TextInputField value={state.lon} label="Breddegrad" onChange={update('lon')} />
+        </>
+    );
+};
+
+export default LatLonInputSet;
--- a/src/components/atoms/LatLonInputSet.tsx
+++ b/src/components/atoms/LatLonInputSet.tsx
@@ -20,8 +20,8 @@ const LatLonInputSet: React.FC<LatLonInputSetProps> = (props) => {
 
     return (
         <>
-            <TextInputField value={state.lat} label="Lengdegrad" onChange={update('lat')} />
-            <TextInputField value={state.lon} label="Breddegrad" onChange={update('lon')} />
+            <TextInputField value={state.lat} label="Breddegrad" onChange={update('lat')} />
+            <TextInputField value={state.lon} label="Lengdegrad" onChange={update('lon')} />
         </>
     );
 };
--- a/src/components/atoms/MunicipalitySelector.tsx
+++ b/src/components/atoms/MunicipalitySelector.tsx
@@ -1,0 +1,20 @@
+import React from 'react';
+import { SelectField } from 'evergreen-ui';
+
+type Props = {
+    label: string;
+    options: [string, string][];
+    onChange: React.ChangeEventHandler<HTMLSelectElement>;
+};
+
+const MunicipalitySelector: React.FC<Props> = (props) => (
+    <SelectField label={props.label} onChange={props.onChange}>
+        {Object.values(props.options).map(([municipalityKey, municipalityName]) => (
+            <option key={municipalityKey} value={municipalityKey}>
+                {municipalityName}
+            </option>
+        ))}
+    </SelectField>
+);
+
+export default MunicipalitySelector;
--- a/src/components/molecules/DataSourceOptions.tsx
+++ b/src/components/molecules/DataSourceOptions.tsx
@@ -1,0 +1,44 @@
+import React, { Dispatch, SetStateAction, useMemo } from 'react';
+import { Heading, Pane } from 'evergreen-ui';
+import { DataSourceID, DataSourceVariables } from '../../types/DataSource';
+import MunicipalityEditor from './MunicipalityEditor';
+import { SSBPopulationVariables } from '../../queries/populationInNorway';
+import LatLonInputSet from '../atoms/LatLonInputSet';
+import { MetAPIVariables } from '../../queries/metApi';
+
+type DataSourceOptionsProps = {
+    state: DataSourceVariables;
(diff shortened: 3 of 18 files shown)


fredrikbw:
I had to add `"start": "react-scripts --max_old_space_size=4096 start"` to `package.json` to run this locally, otherwise running `yarn start` would crash saying  the js heap was out of memory, any idea why @FredrikAugust? It seems to run fine on netlify etc.

src/components/molecules/MunicipalityEditor.tsx
```diff
@@ -0,0 +1,40 @@
...
+    if (error) {
+        return <Text>{error.message}</Text>;
+    }
+
+    if (loading) {
+        return <Combobox isLoading disabled items={['Loading']} />;
+    }
+
+    if (!data) {
+        return <Text>Empty</Text>;
```
blauks:
```suggestion
        return <Text>Tomt</Text>;
```

FredrikAugust:
> I had to add `"start": "react-scripts --max_old_space_size=4096 start"` to `package.json` to run this locally, otherwise running `yarn start` would crash saying the js heap was out of memory, any idea why @FredrikAugust? It seems to run fine on netlify etc.

No clue. Some googling revealed it _might_ be a windows-only problem (?)

src/components/atoms/LatLonInputSet.tsx
```diff
@@ -0,0 +1,29 @@
...
+        props.setState({ lat: Number(state.lat), lon: Number(state.lon) });
+    }, [state]);
+
+    const update = (key: keyof typeof state) => (e: ChangeEvent<HTMLInputElement>) =>
+        setState((prevState) => ({ ...prevState, [key]: e.target.value }));
+
+    return (
+        <>
+            <TextInputField value={state.lat} label="Lengdegrad" onChange={update('lat')} />
+            <TextInputField value={state.lon} label="Breddegrad" onChange={update('lon')} />
```
blauks:
```suggestion
            <TextInputField value={state.lat} label="Breddegrad" onChange={update('lat')} />
            <TextInputField value={state.lon} label="Lengdegrad" onChange={update('lon')} />
```
:))

src/components/molecules/DataSourceOptions.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { DataSourceID, DataSourceVariables } from '../../types/DataSource';
+import MunicipalityEditor from './MunicipalityEditor';
+import { SSBPopulationVariables } from '../../queries/populationInNorway';
+import LatLonInputSet from '../atoms/LatLonInputSet';
+import { MetAPIVariables } from '../../queries/metApi';
+
+type DataSourceOptionsProps = {
+    state: DataSourceVariables;
+    setState: Dispatch<SetStateAction<DataSourceVariables>>;
+
```
blauks, line 12:
```suggestion
```

src/components/molecules/DataSourceOptions.tsx
```diff
@@ -0,0 +1,44 @@
...
+                );
+        }
+    }, [props]);
+
+    return (
+        <Pane gridColumn="span 2">
+            <Heading size={500} marginBottom="1em">
+                Parametere
+            </Heading>
+
```
blauks, line 38:
```suggestion
```

blauks (dismissed):
Looks nice:)))

src/components/atoms/MunicipalitySelector.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { SelectField } from 'evergreen-ui';
+
+type Props = {
```
Kpaubert, line 4: I think we should add a nullable prop for selected municipality in case the user clicks the edit button from the Dashboard, so the correct municipality is pre-selected.

src/components/atoms/LatLonInputSet.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { MetAPIVariables } from '../../queries/metApi';
+import { TextInputField } from 'evergreen-ui';
+
+type LatLonInputSetProps = {
+    state: MetAPIVariables;
+    setState: Dispatch<SetStateAction<MetAPIVariables>>;
+};
+
+const LatLonInputSet: React.FC<LatLonInputSetProps> = (props) => {
+    const [state, setState] = useState<{ lat: string; lon: string }>({ lat: '63', lon: '11' });
```
Kpaubert, line 11: Add props to use pre-selected values in case the user navigated from the edit button in the Dashboard.

Kpaubert (dismissed):
Couple of QoL changes

src/components/atoms/LatLonInputSet.tsx
FredrikAugust: This is part of #126 and #131

src/components/atoms/MunicipalitySelector.tsx
FredrikAugust: This is part of #126 and #131

src/components/molecules/DataSourceOptions.tsx
FredrikAugust: It is intentional to separate the two segments.

src/components/molecules/DataSourceOptions.tsx
FredrikAugust: It is intentional to separate the two segments (state+setState from other props).

blauks (approved):
Coldplay is my favorite band

PR #148, opened 2021-04-21
Add origin of data

### LINKED ISSUE #140: Show origin of data on data explorer page
### Description
Show origin of data (source on MetadataEntry).

Perhaps confer with @IT2901-SINTEF01/ux-design for a quick design mockup. 

### Subtasks
- [x] Show source of data on data explorer page
- [x] Update storybook and other relevant files

### LINKED ISSUE #149: Cypress test suite fails due to application error
### Description
This might be due to the IDs all being the same when testing with cypress.

A better implementation would be to return e.g. two data sources that have their own ID. It had to be changed as the application doesn't work when the ID is random on every call.

### Steps to reproduce
1. Run test suite a few times. Appears it does not _always_ crash

### Expected behaviour
Shouldn't crash.

### Additional comments
Check out the `master` pipeline

blauks:
Closes #140 
(Might also) Closes #149 

Added source URL, data published and data updated to the datasetinfobox:)

### DIFF
--- a/src/components/atoms/DatasetInfoBox.tsx
+++ b/src/components/atoms/DatasetInfoBox.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { Pane, Heading, Text, Badge, Link, Strong } from 'evergreen-ui';
-import { format, formatDistanceToNow } from 'date-fns';
+import { format, formatDistanceToNow, isEqual } from 'date-fns';
 
 type DatasetInfoBoxProps = {
     title: string;
@@ -15,8 +15,31 @@ const DatasetInfoBox: React.FC<DatasetInfoBoxProps> = ({ title, description, tag
     const updatedDate = new Date(updated);
     const publishedDate = new Date(published);
 
+    const Published = () => {
+        return (
+            <Text title={`Published: ${format(publishedDate, 'dd/MM/yyyy;HH:mm')}`} size={300} marginTop="0.15rem">
+                Published: <Strong size={300}>{formatDistanceToNow(publishedDate, { addSuffix: true })}</Strong>
+            </Text>
+        );
+    };
+
+    const LastUpdated = () => {
+        return (
+            <Text
+                title={`Published: ${format(publishedDate, 'dd/MM/yyyy;HH:mm')}, Updated: ${format(
+                    updatedDate,
+                    'dd/MM/yyyy;HH:mm',
+                )}`}
+                size={300}
+                marginTop="0.15rem"
+            >
+                Last updated: <Strong size={300}>{formatDistanceToNow(updatedDate, { addSuffix: true })}</Strong>
+            </Text>
+        );
+    };
+
     return (
-        <Pane width="100%" height="100%">
+        <Pane width="100%" height="100%" display="flex" flexDirection="column">
             <Heading size={600} marginBottom="1rem">
                 {title}
             </Heading>
@@ -39,16 +62,14 @@ const DatasetInfoBox: React.FC<DatasetInfoBoxProps> = ({ title, description, tag
                     </Badge>
                 ))}
             </Pane>
-            <Pane marginTop="1rem" display="flex" flexDirection="column">
+            <Pane marginTop="auto" display="flex" flexDirection="column">
                 <Text size={300}>
-                    URL: <Link size={300}>{url}</Link>
+                    URL:{' '}
+                    <Link href={url} size={300}>
+                        {url}
+                    </Link>
                 </Text>
-                <Text size={300} marginTop="0.15rem">
-                    Last updated: <Strong size={300}>{formatDistanceToNow(updatedDate)}</Strong>
-                </Text>
-                <Text size={300} marginTop="0.15rem">
-                    Published: <Strong size={300}>{format(publishedDate, 'dd/MM/yyyy')}</Strong>
-                </Text>
+                {isEqual(updatedDate, publishedDate) ? <Published /> : <LastUpdated />}
             </Pane>
         </Pane>
     );
--- a/src/components/atoms/DatasetInfoBox.tsx
+++ b/src/components/atoms/DatasetInfoBox.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { Pane, Heading, Text, Badge, Link, Strong } from 'evergreen-ui';
+import { format, formatDistanceToNow } from 'date-fns';
 
 type DatasetInfoBoxProps = {
     title: string;
@@ -43,16 +44,10 @@ const DatasetInfoBox: React.FC<DatasetInfoBoxProps> = ({ title, description, tag
                     URL: <Link size={300}>{url}</Link>
                 </Text>
                 <Text size={300} marginTop="0.15rem">
-                    Last updated:{' '}
-                    <Strong size={300}>
-                        {`${updatedDate.getDate()}.${updatedDate.getMonth() + 1}.${updatedDate.getFullYear()}`}
-                    </Strong>
+                    Last updated: <Strong size={300}>{formatDistanceToNow(updatedDate)}</Strong>
                 </Text>
                 <Text size={300} marginTop="0.15rem">
-                    Published:{' '}
-                    <Strong size={300}>{`${publishedDate.getDate()}.${
-                        publishedDate.getMonth() + 1
-                    }.${publishedDate.getFullYear()}`}</Strong>
+                    Published: <Strong size={300}>{format(publishedDate, 'dd/MM/yyyy')}</Strong>
                 </Text>
             </Pane>
         </Pane>
--- a/src/components/atoms/DatasetInfoBox.tsx
+++ b/src/components/atoms/DatasetInfoBox.tsx
@@ -63,9 +63,7 @@ const DatasetInfoBox: React.FC<DatasetInfoBoxProps> = ({ title, description, tag
                 ))}
             </Pane>
             <Pane marginTop="auto" display="flex" flexDirection="column">
-                <Text size={300}>
-                    URL:{' '}
-                    <Link href={url} size={300}>
+                    <Text size={300}>Origin: </Text>
                         {url}
                     </Link>
                 </Text>
--- a/src/components/atoms/DatasetInfoBox.tsx
+++ b/src/components/atoms/DatasetInfoBox.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { Pane, Heading, Text, Badge, Link, Strong } from 'evergreen-ui';
-import { format, formatDistanceToNow, isEqual } from 'date-fns';
+import { formatISO, formatDistanceToNow, isEqual } from 'date-fns';
 
 type DatasetInfoBoxProps = {
     title: string;
@@ -17,7 +17,7 @@ const DatasetInfoBox: React.FC<DatasetInfoBoxProps> = ({ title, description, tag
 
     const Published: React.FC = () => {
         return (
-            <Text title={`Published: ${format(publishedDate, 'dd/MM/yyyy;HH:mm')}`} size={300} marginTop="0.15rem">
+            <Text title={`Published: ${formatISO(publishedDate)}`} size={300} marginTop="0.15rem">
                 Published: <Strong size={300}>{formatDistanceToNow(publishedDate, { addSuffix: true })}</Strong>
             </Text>
         );
@@ -26,10 +26,7 @@ const DatasetInfoBox: React.FC<DatasetInfoBoxProps> = ({ title, description, tag
     const LastUpdated: React.FC = () => {
         return (
             <Text
-                title={`Published: ${format(publishedDate, 'dd/MM/yyyy;HH:mm')}, Updated: ${format(
-                    updatedDate,
-                    'dd/MM/yyyy;HH:mm',
-                )}`}
+                title={`Published: ${formatISO(publishedDate)}, Updated: ${formatISO(updatedDate)}`}
                 size={300}
                 marginTop="0.15rem"
             >
--- a/src/components/atoms/DatasetInfoBox.tsx
+++ b/src/components/atoms/DatasetInfoBox.tsx
@@ -35,7 +35,7 @@ const DatasetInfoBox: React.FC<DatasetInfoBoxProps> = ({ title, description, tag
         );
     };
 
-    const HandleLinkClick = (e: React.MouseEvent) => {
+    const handleLinkClick: React.MouseEventHandler<HTMLAnchorElement> = (e) => {
         e.stopPropagation();
     };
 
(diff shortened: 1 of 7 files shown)


src/components/atoms/DatasetInfoBox.tsx
```diff
@@ -32,6 +38,23 @@ const DatasetInfoBox: React.FC<DatasetInfoBoxProps> = ({ title, description, tag
...
                 ))}
             </Pane>
+            <Pane marginTop="1rem" display="flex" flexDirection="column">
+                <Text size={300}>
+                    URL: <Link size={300}>{url}</Link>
+                </Text>
+                <Text size={300} marginTop="0.15rem">
+                    Last updated:{' '}
+                    <Strong size={300}>
+                        {`${updatedDate.getDate()}.${updatedDate.getMonth() + 1}.${updatedDate.getFullYear()}`}
```
FredrikAugust: Date-fns has support for a variety of formatting methods, including ["distance to now in words"](https://date-fns.org/v2.21.1/docs/formatDistanceToNow) which might be a bit more human friendly. It is a common pattern to have the date represented as "4 years ago", and when you hover it you get a tooltip of the exact date.

Furthermore, by using date-fns you could format based on the users locale, which improves user experience.

src/components/molecules/DataResultItem.tsx
```diff
@@ -8,13 +8,24 @@ export type DataResultItemProps = {
...
 
-const DataResultItem: React.FC<DataResultItemProps> = ({ title, description, tags, visualisationType }) => (
+const DataResultItem: React.FC<DataResultItemProps> = ({
+    title,
+    description,
+    tags,
+    visualisationType,
+    source,
+    updated,
+    published,
```
FredrikAugust, line 24: Might be easier to pass down the `metadata` object instead and specify what fields you require with 

```ts
type Props = Pick<MetadataEntry, 'title' | 'description' | ...>;
// or conversely Omit<MetadataEntry, '...' | '...' >;
```

That way you could create the DataResultItem as `<DataResultItem {...metadataEntry} />`

No need to change this now if you don't feel like it, but worth keeping in mind for later to avoid repetition.

FredrikAugust (changes requested):

src/components/atoms/DatasetInfoBox.tsx
FredrikAugust: As we've already included it as a dep I think we should use it's pwoers.

src/components/molecules/DataResultItem.tsx
blauks: Good tip, I will keep that in mind! But I will keep it unchanged for now:):):)

src/components/atoms/DatasetInfoBox.tsx
blauks: That's a better solution, I will change it 🤠7

src/components/atoms/DatasetInfoBox.tsx
```diff
@@ -32,6 +39,17 @@ const DatasetInfoBox: React.FC<DatasetInfoBoxProps> = ({ title, description, tag
                     </Badge>
                 ))}
             </Pane>
+            <Pane marginTop="1rem" display="flex" flexDirection="column">
+                <Text size={300}>
+                    URL: <Link size={300}>{url}</Link>
+                </Text>
+                <Text size={300} marginTop="0.15rem">
+                    Last updated: <Strong size={300}>{formatDistanceToNow(updatedDate)}</Strong>
```
FredrikAugust: Include suffix

https://date-fns.org/

It's on the front page

https://date-fns.org/

FredrikAugust: Saying a data source was published "5 months" doesn't say much

FredrikAugust (changes requested):
- The URL doesn't appear to work (Firefox) as the link goes to the edit vis. page instead
- Could the extra information be aligned to the bottom of the box? Separated from the tags and description? Should be simple with a flexbox. This would perhaps look a bit more refined and visually balanced
- Do we need to display updated and published if they're the same date?
- Can both perhaps be displayed in the same format? 
- Could you for example _only_ display when it was last updated in relative format, and display the exact date of updating/publishing in a [title](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title)?

src/components/atoms/DatasetInfoBox.tsx
blauks: Oh oops, I will fix that. Could you send the link again?

blauks:
> The URL doesn't appear to work (Firefox) as the link goes to the edit vis. page instead

Forgot href :):):)):):)

> Could the extra information be aligned to the bottom of the box? Separated from the tags and description? Should be simple with a flexbox. This would perhaps look a bit more refined and visually balanced

It could. I do not know what you mean with "refined and visually balanced", but I can try it out and see if it looks better.

> Do we need to display updated and published if they're the same date?

Nope. 

> Could you for example only display when it was last updated in relative format, and display the exact date of updating/publishing in a title?

Good idea, I think that would be cool 😎. So basically just keep the layout, but add a title to last updated?

src/components/atoms/DatasetInfoBox.tsx
FredrikAugust: Yes of course sorry mister

https://date-fns.org/

FredrikAugust:
> > The URL doesn't appear to work (Firefox) as the link goes to the edit vis. page instead
> 
> Forgot href :):):)):):)
> 
> > Could the extra information be aligned to the bottom of the box? Separated from the tags and description? Should be simple with a flexbox. This would perhaps look a bit more refined and visually balanced
> 
> It could. I do not know what you mean with "refined and visually balanced", but I can try it out and see if it looks better.
> 
> > Do we need to display updated and published if they're the same date?
> 
> Nope.
> 
> > Could you for example only display when it was last updated in relative format, and display the exact date of updating/publishing in a title?
> 
> Good idea, I think that would be cool 😎. So basically just keep the layout, but add a title to last updated?

Regarding visual balance; I meant that it looks slightly _imbalanced_ when everything is aligned to the top. By using the entire height of the card, you can distribute the "visual weight" (basically just amount of content and how easy you notice it) and make it feel more balanced and legible. If you want to read more about it, I can recommend the book "Don't make me think", and [this article](https://www.smashingmagazine.com/2014/12/design-principles-visual-weight-direction/) appears to give a good explanation.

And regarding the final question;

If it's been updated, _only_ display the "last updated: 5 months ago", and give it a "title" something like "Published: 23/08/1999. Updated 21/04/2021".

if it hasn't been, _only_ display the "published: 5 months ago" and give it a title along the lines of "Published: 23/08/1999".

blauks:
> > > The URL doesn't appear to work (Firefox) as the link goes to the edit vis. page instead
> 
> > 
> 
> > Forgot href :):):)):):)
> 
> > 
> 
> > > Could the extra information be aligned to the bottom of the box? Separated from the tags and description? Should be simple with a flexbox. This would perhaps look a bit more refined and visually balanced
> 
> > 
> 
> > It could. I do not know what you mean with "refined and visually balanced", but I can try it out and see if it looks better.
> 
> > 
> 
> > > Do we need to display updated and published if they're the same date?
> 
> > 
> 
> > Nope.
> 
> > 
> 
> > > Could you for example only display when it was last updated in relative format, and display the exact date of updating/publishing in a title?
> 
> > 
> 
> > Good idea, I think that would be cool 😎. So basically just keep the layout, but add a title to last updated?
> 
> 
> 
> Regarding visual balance; I meant that it looks slightly _imbalanced_ when everything is aligned to the top. By using the entire height of the card, you can distribute the "visual weight" (basically just amount of content and how easy you notice it) and make it feel more balanced and legible. If you want to read more about it, I can recommend the book "Don't make me think", and [this article](https://www.smashingmagazine.com/2014/12/design-principles-visual-weight-direction/) appears to give a good explanation.
> 
> 
> 
> And regarding the final question;
> 
> 
> 
> If it's been updated, _only_ display the "last updated: 5 months ago", and give it a "title" something like "Published: 23/08/1999. Updated 21/04/2021".
> 
> 
> 
> if it hasn't been, _only_ display the "published: 5 months ago" and give it a title along the lines of "Published: 23/08/1999".

Sounds nice and cool

src/components/atoms/DatasetInfoBox.tsx
blauks: Thanks

src/components/atoms/DatasetInfoBox.tsx
```diff
@@ -32,6 +62,15 @@ const DatasetInfoBox: React.FC<DatasetInfoBoxProps> = ({ title, description, tag
                     </Badge>
                 ))}
             </Pane>
+            <Pane marginTop="auto" display="flex" flexDirection="column">
+                <Text size={300}>
+                    URL:{' '}
```
FredrikAugust: Perhaps a more descriptive name? Source?

src/components/atoms/DatasetInfoBox.tsx
```diff
@@ -1,15 +1,45 @@
...
 };
 
-const DatasetInfoBox: React.FC<DatasetInfoBoxProps> = ({ title, description, tags }) => {
+const DatasetInfoBox: React.FC<DatasetInfoBoxProps> = ({ title, description, tags, url, updated, published }) => {
+    const updatedDate = new Date(updated);
+    const publishedDate = new Date(published);
+
+    const Published: React.FC = () => {
+        return (
+            <Text title={`Published: ${format(publishedDate, 'dd/MM/yyyy;HH:mm')}`} size={300} marginTop="0.15rem">
```
FredrikAugust: Use [`formatISO`](https://date-fns.org/v2.21.1/docs/formatISO) which is a universal standard. That includes timezones and can be easily parsed back and interpreted.

src/components/atoms/DatasetInfoBox.tsx
```diff
@@ -32,6 +62,15 @@ const DatasetInfoBox: React.FC<DatasetInfoBoxProps> = ({ title, description, tag
                     </Badge>
                 ))}
             </Pane>
+            <Pane marginTop="auto" display="flex" flexDirection="column">
+                <Text size={300}>
+                    URL:{' '}
+                    <Link href={url} size={300}>
+                        {url}
+                    </Link>
```
FredrikAugust: This still doesn't work with a single click.
Tested on Firefox.

src/components/atoms/DatasetInfoBox.tsx
FredrikAugust: - Origin?

src/components/atoms/DatasetInfoBox.tsx
FredrikAugust: > Rekkefølgen år-måned-dag er norsk og internasjonal standard og mye brukt ved utveksling av data, i kronologiske lister m.m.

https://www.sprakradet.no/sprakhjelp/Skriveregler/Dato/#dato

Dette er en del av ISO-8601

FredrikAugust (changes requested):
Most crucial part is that the link should work

src/components/atoms/DatasetInfoBox.tsx
FredrikAugust: Does work with a "Right click -> Open in new tab".

I suspect someone put an `onClick` handler on the cards which is interfering with everything else that wants to be clicked on inside it. Or perhaps a Link surrounding it.

You might need to 

a) change the onClick/Link on the card
b) do some trickery with event propagation and bubbling

FredrikAugust: Doesn't work in Safari either.

FredrikAugust: I'd appreciate it if you could test this properly before re-requesting a review.

blauks: It worked on my firefox. I tested it before I pushed. I will look more into the problem

FredrikAugust: By the way, I am referring to the data explorer links. I.e. not on the `/explore/edit/...` page, but on `/explore`. Did those work on your firefox as well?

FredrikAugust:
Screenshot 2021-04-22 at 11 11 33

It might benefit from some padding here as well between the tags and the "Updated" field.

_Note: this only happens with a decently sized description field. Perhaps some margin to the bottom of the description would do._

src/components/atoms/DatasetInfoBox.tsx
blauks: My bad, I will fix it.

src/components/atoms/DatasetInfoBox.tsx
blauks: Okey doke

Kpaubert (dismissed):
LGTM

src/components/atoms/DatasetInfoBox.tsx
```diff
@@ -38,6 +35,10 @@ const DatasetInfoBox: React.FC<DatasetInfoBoxProps> = ({ title, description, tag
         );
     };
 
+    const HandleLinkClick = (e: React.MouseEvent) => {
```
FredrikAugust: Small first letter in function/variable names

FredrikAugust:
```suggestion
    const handleLinkClick: React.MouseEventHandler<HTMLAnchorElement> = (e) => {
```

src/components/atoms/DatasetInfoBox.tsx
```diff
@@ -63,12 +64,12 @@ const DatasetInfoBox: React.FC<DatasetInfoBoxProps> = ({ title, description, tag
                 ))}
             </Pane>
             <Pane marginTop="auto" display="flex" flexDirection="column">
-                <Text size={300}>
-                    URL:{' '}
-                    <Link href={url} size={300}>
+                <Pane>
+                    <Text size={300}>Origin: </Text>
+                    <Link href={url} target="_blank" onClick={(e: React.MouseEvent) => HandleLinkClick(e)} size={300}>
```
FredrikAugust:
```suggestion
                    <Link href={url} target="_blank" onClick={handleLinkClick} size={300}>
```

src/components/organisms/DataResultItems.tsx
```diff
@@ -46,21 +47,31 @@ const DataResultItems: React.FC<DataResultItemsProps> = ({ activeFilters }) => {
         );
     }
 
+    const HandleItemClick = (id: string) => {
```
FredrikAugust: No capital first letter

src/components/organisms/DataResultItems.tsx
```diff
@@ -46,21 +47,31 @@ const DataResultItems: React.FC<DataResultItemsProps> = ({ activeFilters }) => {
...
-                        />
-                    </Link>
+                <Pane
+                    width="90%"
+                    margin="auto"
+                    marginTop="2rem"
+                    key={el.id}
+                    backgroundColor="white"
+                    onClick={() => HandleItemClick(el.id)}
+                    style={{ cursor: 'pointer' }}
```
FredrikAugust: Use the `cursor` prop on `Pane`.

FredrikAugust (changes requested):

src/components/atoms/DatasetInfoBox.tsx
FredrikAugust: Have you been looking too much at the backend?;)

blauks: :D

FredrikAugust (approved):

PR #152, opened 2021-04-22
Hydrate edit page with variables

### LINKED ISSUE #126: Hydrate edit page with parameters from visualisation

FredrikAugust:
Closes #126 

------------

Changelog
------
- The edit page is now seeded with the values from the Redux store
- Moved effect that followed an if-statement causing errors on page refresh due to differing amount of hook invocations

### DIFF
--- a/src/components/atoms/LatLonInputSet.tsx
+++ b/src/components/atoms/LatLonInputSet.tsx
@@ -8,7 +8,10 @@ type LatLonInputSetProps = {
 };
 
 const LatLonInputSet: React.FC<LatLonInputSetProps> = (props) => {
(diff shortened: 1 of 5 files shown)


Kpaubert (changes requested):
Nothing to say on the code, but the description should also be hydrated. If the user looks at a specific visualization (one that is added to the dashboard), the description could be updated on the fly while the user edits it - without having to click "Remove from dashboard" and then re-add it.

Kpaubert (approved):
Approving this, putting the review on #153

